PR #80, opened 2015-01-27
Use $(MAKE) instead of make because make can be `gmake`, `make4`, etc.

KeenS:


drmeister:
Got it - thank you very much!

PR #95, opened 2015-05-01
Switch List_sp_iterator to use Cons_sp

Georgiy-Tugai:
Builds, but doesn't link (`multiple definitions of ... operator==`)

We might have to make fast_iterator an actual seperate class, rather than template hackery (dummy bool template parameter) to make the linker happy, not sure.


drmeister:
Got it

PR #96, opened 2015-05-02
Get rid of isNil in Nilable<_sp>

Georgiy-Tugai:


drmeister:
Got it!  Thank you very much!

PR #192, opened 2015-12-05
Add an archive task to generate a source package

bendlas:
Also add a build_all task, to build w/o updating submodules from the
exported package. The (default) all task behaves as before.

The archive export is achieved by a script `git_archive_all.py` from
https://github.com/Kentzo/git-archive-all

.keep files are added in `generated` dirs, so that they are guaranteed
to be present in the source package.


Georgiy-Tugai:
Please rebase this onto `dev`; the `master` branch is intended for releases.

I am currently working on (in the `dev-makefile` branch) making the Makefile significantly more declarative, which should remove the need for a `build_all` target -- submodule fetches will be conditional on the relative ages of key files in the submodule directories and `.gitmodules`.

bendlas:
OK, I thought `release` was for releases. I rebased now, unfortunately, I can't change the target branch of the PR, but you should be able to merge cleanly into `dev`, locally.

Shinmera:
This is wildly outdated.

PR #229, opened 2016-03-06
Optimize #x=/#x#.

stassats:
Instead of using a symbol as a tag marking yet unread objects use a
structure.
It is more compact, exactly identifiable as a tag and can store the
read value directly instead of using an additional list.


drmeister:
I'm changing the code so that it doesn't use the CL struct SHARP-TAG but rather uses the C++ class CORE:SHARP-EQUAL-WRAPPER.    It's a drop in replacement and the change can be made by removing the (defstruct sharp-tag ...) and using  s/sharp-tag/core:sharp-equal-wrapper/g

drmeister:
Of course I had to fix make-sharp-equal-wrapper after doing the substitution above - but now it should be all consistent.

PR #231, opened 2016-03-10
Documentation of the bootstrap process

rptb1:
I've written up the notes from our conversation today into a maintainable document.

I suggest cross-referencing it from the source code. At Ravenbrook we use tags -- see for example the string “design.mps.splay.prop” in https://github.com/Ravenbrook/mps/blob/master/code/splay.c . These kinds of cross references have survived 20 years of change and several source control systems!

This is an update to #230 with corrected branching!


rptb1:
@drmeister How does SBCL fit into the bootstrap process?

Shinmera:
SBCL runs the scraper parts.

rptb1:
@Shinmera Thanks for https://github.com/drmeister/clasp/issues/182 .  I'll get an overview direct from @drmeister and include details from that, and insert a reference to it.  Do you know of any other docs (in issues etc.) that can be usefully referenced from bootstrap.rst ?

Shinmera:
I don't think so. Outside of the work done to clean up things such as in the makefiles, the bootstrap process has not been approached by anyone.

drmeister:
I've held off merging this pull request because I rewrote the build system using 'waf' and some things have changed although the document is largely correct.

Should I merge this and edit the file or edit the file somehow before merging?

Shinmera:
As long as you don't merge it into testing directly as it is here I don't really care. The changes seem rather minimal.

Shinmera:
Merged as #304

PR #255, opened 2016-06-24
CHANGED/FIXED:  (in C++ land) Bignum output as string fixed on compil…

dg1sbg:
…e level - STILL NEEDS TESTING.


Shinmera:
Currently in this patch braces are sometimes on their own lines, which is not in conformance with the style of the rest of the codebase. Please resubmit the PR with patches for a consistent style.

PR #256, opened 2016-06-24
CHANGED/FIXED:  (in C++ land) Bignum output as string fixed. Added -lgmp to link flags.

dg1sbg:


src/core/bignum.cc
```diff
@@ -58,6 +58,24 @@ CL_DEFUN Bignum_sp Bignum_O::make(const string &value_in_string) {
   return ((bn));
 };
 
+string Bignum_O::valueAsString() const {
+  stringstream      ss;
+  int               ibase = 10;
+  char            * buffer = NULL;
+
+  buffer = mpz_get_str( NULL, ibase, this->_value.get_mpz_t() );
```
drmeister: Does this leak memory?   Perhaps we should have a predefined buffer to mpz_get_str into

include/clasp/core/externalObject.h
```diff
@@ -1,38 +1,78 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+// ===========================================================================
+
+//! @file   externalObject.h
+//! @author Christian E. Schafmeister
+//! @brief
```
Georgiy-Tugai: Is all this header stuff really necessary? The first line somewhat makes sense, although we have `.dir-locals.el` for that... but there's no need for the figlet text, or the doxygen-style stuff for that matter, unless @drmeister wants to use doxygen.

Admittedly, I don't know enough about the legal aspects to be sure whether the disclaimer/copyright is in fact needed in every file. Maybe it is.

Shinmera: Including the notice is not necessary. A simple LICENSE file at the root should suffice. Adding a note to every file has become widespread because the GPL "advises" it.

Either way we should stick to the notice we already have in the files. Something new like this, especially the ASCII graphic, is just needless crap in the files that nobody wants to see.

dg1sbg: To each his own. These are my changes and I need them. If you don’t want them you are most welcome top not look at that „crap“.

Am 25.06.2016 um 11:12 schrieb Nicolas Hafner <notifications@github.com>:

Shinmera: What do you *need* an ASCII graphic header for?

Either we're going to have a consistent source style throughout the repository, or I'm going to resign from all responsibilities I've taken up for this project.

Georgiy-Tugai: @dg1sbg For future reference, I recommend separating major comment changes (such as your file headers), whitespace-only changes (e.g. re-indenting) and re-ordering things into clearly labelled independent commits, separate from the "body" of your work; it's a lot easier to review -- and, if considered necessary, cherry-pick -- things that way.

There are many ways to stage hunks (or individual line changes) independently; examples include [git's own Interactive Staging](https://git-scm.com/book/en/v2/Git-Tools-Interactive-Staging) and porcelains such as [the `magit` Emacs extension](https://magit.vc/).

PR #261, opened 2016-07-13
Add CONTRIBUTING document

wemeetagain:
Add `CONTRIBUTING`, `PULL_REQUEST_TEMPLATE`, and `ISSUE_TEMPLATE`
See:
https://irclog.tymoon.eu/freenode/clasp?around=2016-07-12T17:27:36&types=mnaot#1468344456
https://github.com/blog/1184-contributing-guidelines
https://github.com/blog/2111-issue-and-pull-request-templates
Note: The templates must be added to the default branch to take effect.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,6 @@
+### Expected behavior
+
+### Actual behavior
+
+### Steps to reproduce the behavior
+
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,6 @@
+Fixes # .
+
+Changes proposed in this pull request:
+- 
+- 
+- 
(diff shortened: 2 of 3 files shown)


mhsjlw:
Hmm, would it also be a good idea to write a code of conduct file? It's a big project, I don't think it'd hurt if we had one.

PuercoPop:
Adding an issue and pull request template would be a good idea as well https://github.com/blog/2111-issue-and-pull-request-templates

wemeetagain:
@mhsjlw I would defer to someone else's judgment. I really sent in this PR to pin down the details of the referenced chat log somewhere that wouldn't get lost.

Georgiy-Tugai:
@Shinmera @drmeister opinions?

Since these files must be added to the default branch, we'll either have to make `dev` default, or violate the `master` policy to add them there.

Shinmera:
Pushing commits to master is something that is necessary occasionally for things like hotfixes or documentation updates that cannot come from the dev upstream. Thus I'm fine with adding these files, but perhaps not with the contents that are suggested here.

I'll see what I can do about this later today, right this moment I'm busy with other things.

mhsjlw:
@wemeetagain yup, makes sense. I'm in the wrong place for this.

drmeister:
I'm pulling this into dev and then migrating it into testing and later preview/master

PR #266, opened 2016-08-13
PR: FRGO_CLASP_FOREIGN_DATA_A0100

dg1sbg:


src/lisp/kernel/init.lsp
```diff
@@ -1005,7 +1005,9 @@ Return files."
          (bformat t "Starting %s ... loading image... it takes a few seconds\n" (lisp-implementation-version)))))
 
 
+(export '*extension-startup-loads*) ;; ADDED: frgo, 2016-08-10
```
Shinmera: Don't we have version control to keep track of what was changed when?

drmeister:
It is done.

PR #289, opened 2016-09-02
Fix various

dkochmanski:
This fixes a few problems encountered during work on ansi-tests for clasp. Build is clean, quicklisp + swank compile without problems.


dkochmanski:
not this branch

PR #290, opened 2016-09-02
Fix a Variety of Issues Preventing the Ansi Test Suite from Running

dkochmanski:
This fixes a few problems encountered during work on ansi-tests for clasp. Build is clean, quicklisp + swank compile without problems. Commits on top of dev-ecl branch.


src/lisp/kernel/cmp/compiler.lsp
```diff
@@ -1459,7 +1459,8 @@ We could do more fancy things here - like if cleavir-clasp fails, use the clasp
 	      (generate-lambda-expression-from-interpreted-function func)
 	    (cmp-log "About to compile  name: %s  lambda-expression: %s wrapped-env: %s\n" name lambda-expression wrapped-env)
 	    (compile-in-env name lambda-expression wrapped-env compile-hook)))
-	 (t (error "Could not compile func")))))
```
Shinmera: Could you elaborate on why it was necessary to avoid an error in this case?

src/lisp/kernel/lsp/arraylib.lsp
```diff
@@ -354,6 +407,22 @@ adjustable array."
     (sys:replace-array array x)
     ))
 
+#+clasp
+(defun adjust-array (array dimensions &key element-type initial-element initial-contents fill-pointer displaced-to displaced-index-offset)
+  (and fill-pointer (error "Add support for fill-pointers in arrays"))
```
Shinmera: Wait, Clasp doesn't have support for fill-pointers yet?

Shinmera:
Aside from the one comment that I'm not sure about, this looks good & clean to me. Thanks for the few style improvements in some places as well.

drmeister:
Ok, I think we are done here.  jackdaniel's changes are merged.

PR #308, opened 2016-10-18
Build cleanup

attila-lendvai:
## major changes

- Invoking ./waf directly should work now in a clean shell
- New python based config file
- Deleted the scraper shell scripts and call SBCL directly

## known issues
- installation adds an extra 'fasl' directory level that is not expected by cclasp, and which breaks `(require :asdf)` (i think it's not a regression)


drmeister:
I made some additional changes to the wscript file to add debugging options for build commands.

PR #309, opened 2016-10-19
Simplify wscript.config format

jorams:
Note that after this the only thing you can do in wscript.config is to
set keys in the configuration context's `env' property.

### DIFF
--- a/wscript
+++ b/wscript
@@ -428,8 +428,9 @@ def configure(cfg):
         if not os.path.isfile("./wscript.config"):
             print("Please provide the required config for the build; see the wscript.config.template file.")
             sys.exit(1)
-        exec(open("./wscript.config").read(), globals(), locals())
-        configure_local(cfg)
+        local_environment = {}
+        exec(open("./wscript.config").read(), globals(), local_environment)
+        cfg.env.update(local_environment)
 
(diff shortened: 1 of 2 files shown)


jorams:
Cancelling this. `cfg.env` is not as simple of a dictionary as I thought it was, and some of the functionality it provides is useful.

jorams:
After some more research, I think this could be merged after all. At the point where this file is loaded, nothing of significance is stored in `cfg.env` yet, so the additional methods are of no use.

Sorry for the spam.

Shinmera:
Hmm. I'm not so sure if there shouldn't be a hybrid solution.

Something like a standard `config.py` which offers the same capabilities as the current `wscript.config`, but by default delegates to loading a simple dictionary from `parameters.conf` (or maybe even a `.ini`?) might be better.

attila-lendvai:
i don't see why it's helpful to constrain the python universe in the config file.

if the user just wants to set variables, then the examples are there in comments. if he wants more, he can do what python allows.

at one point it may even be extended with more hooks that are called at different stages of the build process.

drmeister:
Rebased and merged

Shinmera:
Once again a merge is in without properly discussing things first.

Sigh.

PR #311, opened 2016-10-23
fix up CL:COMPILE a little

### LINKED ISSUE #300: COMPILEing An Already Compiled Function Signals a Warning
I don't think this is permitted.

```
> (defun foo () 0)

FOO
> (compile 'foo)
;;; Warning: We have lost the original function definition for #<CLOSURE-WITH-SLOTS FOO :ftype :FUNCTION lambda-list: COMMON-LISP:NIL :fptr 0x7fa9749c42f0>. Compilation failed.

#<CLOSURE-WITH-SLOTS FOO :ftype :FUNCTION lambda-list: COMMON-LISP:NIL :fptr 0x7fa9749c42f0>
T
NIL
```

Bike:
Specifically, it should now return three values, and knows how to deal
with definitions that are functions.

Should fix #300


Shinmera:
Please don't patch and PR against `testing`. And @drmeister please never merge PRs against the wrong branch.

PR #320, opened 2016-12-06
build: clean up unicode/string headaches, factor out run_program

attila-lendvai:
please merge with rebase (and not by squashing it).

### DIFF
--- a/wscript
+++ b/wscript
@@ -429,35 +429,36 @@ def options(cfg):
     cfg.load('compiler_cxx')
     cfg.load('compiler_c')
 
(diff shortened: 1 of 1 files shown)


attila-lendvai:
the key change was to pass `universal_newlines = True` to `subprocess.Popen`, and remove any `.encode()` and/or `.decode()` calls.

this way every relevant value will be a string (as opposed to `bytes`, or whatever a byte vector is called in python, that behaves a lot like strings, but not completely, for some extra confusion...)

the rest is just a small refactor, to make sure there's only one place that calls `Popen`.

PR #321, opened 2016-12-08
Cleanup

attila-lendvai:
delete stale files and definitions, including workbench.lisp (!) files. if you need them on different machines, then i suggest either moving them under a personal repo, or even better setting up some dropbox-like auto-sharing of a directory under your home.

build has been tested to work.


src/lisp/kernel/init.lsp
```diff
@@ -855,6 +856,7 @@ the stage, the +application-name+ and the +bitcode-name+"
                  `(my-do-time #'(lambda () ,form))))
            t)
 (export 'my-time)
+)
```
Shinmera: I think this should just be removed altogether. It isn't referenced anywhere else.

attila-lendvai: yes, it can be dropped, but @drmeister may want to copy it somewhere else. please confirm that you noted this, and then i'll amend the PR to delete it.

Shinmera: If he really needs it, it'll be in git's history. It's not like it'll be lost forever.

attila-lendvai: fair enough, PR amended.

attila-lendvai:
alternatively, the workbench files could be merged and moved under docs/ because its contents can be useful when communicating with others, too. to serve as something that can be referenced in discussions.

attila-lendvai:
FTR, i'm working on rebasing this PR.

attila-lendvai:
rebased to dev as of now, and has built successfully.

attila-lendvai:
tested to build fine and it starts slime. i think it's ready for merging (use "rebase and merge").

PR #342, opened 2017-04-08
PR for fixing race condition handling during #'ext:vfork-execvp

dg1sbg:
Also includes some fixes for CFFI support.


drmeister:
frgo: I looked over and then incorporated your changes.
Thank you very much!

PR #343, opened 2017-04-10
Testing

benton:
This version uses a feature of Dockerfile's `ADD` directive that automatically unpacks archive files. The build output from clasp is packaged up into a single `./build/clasp.tgz` file, instead of a full directory of files. This more reliably preserves user and group IDs across the build and runtime images, and avoids "littering" the user's workstation unnecessarily.

Running `chown -R app:app $HOME/clasp` within a docker build turns out to generate a new filesystem layer with a full copy of every original file, effectively doubling the size of the layer's contents even though only ownership is modified. To avoid this problem, the `app` user and `chown` statement have been (re-)added to the build image. The `clasp.tgz` file generated by `docker-compose run clasp-build` preserves the ownership of the files when they are `ADD`ed to the runtime image.

The final runtime image for `clasp` is about 1.1GB. When based on the 130MB ubuntu base, this is about as small as it can get, given that the build output itself is over 800MB, with LLVM adding another 40MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,9 +18,7 @@ boost_build_v2/bjam
 wscript.config
 boost_build_v2/engine/bin.linuxx86_64
 boost_build_v2/engine/bootstrap
-# ensure the build directory exists on checkout
 build/*
-!build/.gitkeep
 projects/
 build2/
 /.lock-waf_linux*_build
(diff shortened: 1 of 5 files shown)


drmeister:
I merged this with the command line tools

PR #347, opened 2017-04-18
Dev: Implementation of Queue_O - a Multiple-Producer Multiple-Consumer concurrent queue

dg1sbg:
Necessary basic element for implementing Mailboxes. Implementation of Queue_O is done in the MP package.


include/clasp/core/queue.h
```diff
@@ -0,0 +1,101 @@
...
+
+    CL_LISPIFY_NAME("%count");
+    CL_LAMBDA(queue);
+    CL_DOCSTRING("doc(Returns the number of objects in QUEUE. Mainly useful for manual examination of queue state, and in PRINT-OBJECT methods: inefficient as it must walk the entire queue.)doc");
+    CL_DEFMETHOD core::Integer_sp PERCENTcount( void );
+
+  public: // SLOTS
+
+    core::String_sp                            _Name;
+    moodycamel::ConcurrentQueue< core::T_sp >  _Queue;
```
drmeister: This is going to be a problem.  I'm guessing the core::T_sp objects are going to be allocated on the heap, outside of the garbage collectors pools.   That won't work, they need to be allocated inside the GC managed memory.  All GC managed pointers need to be inside GC managed memory.

We need to figure out how to deal with this.

include/clasp/core/queue.h
```diff
@@ -0,0 +1,101 @@
...
+    core::ThreadLocalState *                   _ThreadInfo;
```
drmeister: Why save the ThreadLocalState* pointer?    It's always available to the thread.

include/clasp/core/queue.h
```diff
@@ -0,0 +1,101 @@
...
+  CL_LAMBDA(&key name);
+  CL_DOCSTRING("doc(Returns a new QUEUE with NAME.)doc");
+  CL_DEFUN Queue_sp PERCENTmake_queue( core::String_sp name );
+
+}; // namespace
+
+// GARBAGE COLLECTION CONFIG
+
+template <>
+struct gctools::GCInfo<mp::Queue_O> {
```
drmeister: This should be moved up to before the Queue_O definition in case in the future we call the allocators from within the header.

PR #348, opened 2017-04-22
SICL changes

Bike:
needs the Bike/SICL clasp-changes branch. No idea how pull requests work with submodules. None.

The unmerged code builds and works on my machine. I haven't built the merged version yet (takes several hours after all) but given the smallness of the only merge conflict I think it should be okay.


Bike:
Builds, runs SLIME.

drmeister:
Merged

PR #355, opened 2017-06-09
Cloudformation

drmeister:
Bringing in Benton's changes to support the setup of CI service on Amazon Cloud


drmeister:
Merge is done.

PR #359, opened 2017-06-27
Fix bug in parallel compilation and remove big compiler mutex

Bike:
LLVM is parallel and so is Cleavir, so for Clasp to do multiple compilations
simultaneously is easy in principle. As far as I can tell, the only remaining
problem was COMPILER::*PRIMITIVES*. This global hash table was refilled by
several entry points to the compiler whenever they made a module, and referenced
thereafter quite a bit. Now that *PRIMITIVES* is rebound instead of modified,
compiling in parallel seems to work.


drmeister:
Done

PR #377, opened 2017-09-14
Initial checkin of Gentoo ebuild

Haifen:
The current (Git live, hence the -9999 version number) ebuild does function as completely as a manual build process, but may still have some circumstances in which it will behave badly or fail (see the README.md for the start of the documentation of uncovered corner cases).

I will try to keep this branch up-to-date and submit pull requests to the [upstream repo](https://github.com/drmeister/clasp) (currently https://github.com/drmeister/clasp) as often as possible (excluding versions of the ebuild that break), but there's likely to be a slightly-faster-moving ebuild at https://github.com/Haifen/robinkverlay/tree/master/dev-lisp/clasp (with the caveat that it's more likely to break).

Direct complaints about the ebuild to [me](https://github.com/Haifen) ([Haifen](https://github.com/Haifen)/[robink@creosotehill.org](mailto:robink@creosotehill.org)) first; at the time of writing this commit message, [drmeister](https://github.com/drmeister) doesn't have easy access to a Gentoo system to test with, so he'll be harder-pressed to track issues down, and breakage is also likely to be my fault, not his.


Shinmera:
While I can't verify this myself as I lack a gentoo system right now, I see no reason to not merge this in either, since it doesn't conflict with any actual development files.

PR #402, opened 2018-01-08
convert simple errors in reader-errors (those used in ansl-cl-tests)

kpoeck:
As discussed in IRC. Convert simple-errors that are really reader-errors. Only touched the things that occured during ansi-cl-testing.

Did try to distinguish errors from top-level (stream, but no file), from the uses in a file. Did test all combinations.

Could add test-cases, if you tell me the framework used


Bike:
src/lisp/regression-tests. it's pretty barebones right now.

Bike:
The comparing with -no-name- isn't great. I mean, I understand why you did it that way, but I feel like we should rather have the function you get the string from return some dedicated option failure value instead of that.

PR #416, opened 2018-02-25
initalize format-error condition better, so that printing does not error

### LINKED ISSUE #407: Poor errors from malformed FORMAT strings
For example `(format nil "~]")` results in `Error (UNBOUND-SLOT) during printing: #<FORMAT-ERROR >`

Apparently because the format-error printer expects things like the actual format string to be in the condition, and format does not properly provide them.

kpoeck:
Fixes #407


Bike:
ef8c198 is basically the same patch, that i did yesterday and built overnight, so this PR is now redundant. Thanks for kicking my ass into gear though.

PR #419, opened 2018-03-10
Build refactor

attila-lendvai:
i reached a point where the code is ready for feedback.

highlights:
- all the build-related code is collected into the wscript file and tools-for-build/
- build output is through a proper logging system. info output goes to the console, debug output is appended to build/variant/build.log (with timestamps)
- build-time options have been turned into --noscrape like command line parameters, properly separated from the switches changing the build output. no more noscrape_cboehm_d targets. only the variant and the debug_char remaned in the build target name (whatever changes the build *output* as opposed to the build run).
- update_submodule should work as expected with local changes; also reminds the user of pending local changes in its output
- pump template expansion has been integrated into the build (**NOTE**: this change removes the `.../generated/...` from the `#include` paths. generated include files are now placed in `build/boehm/generated/clasp/`).
- locale is fixed to "C" to avoid host config affecting the build. scraper runs using utf-8.
- prints the build duration into the log

pending questions:
- now that pump.py is integrated into the build, i think we should delete its output from the repo and generate the .h files into build/variant/generated/ (i'll work on this and update the PR)


attila-lendvai:
as discussed on IRC, i suggest the following order:

 1) optionally: freshly merge dev HEAD into cst to resolve any clasp-internal conflicts
 2) merge this into dev
 3) and then i'll merge dev into cst again and resolve the build related conflicts, test it, and push it to cst

attila-lendvai:
dev diverged and i had to merge/rebase the PR. it would be nice to avoid further conflicts.

@drmeister if you give the green light on this, then i can deal with merging it into dev, test the build, and then merge dev into cst.

PR #446, opened 2018-03-17
First attempt to set up travis based on a new Dockerfile.

drmeister:
Unfortunately the current max time limit in travis is 50 mins, which is
nowhere near enough for clasp to build.

I just purchased a one year travis-ci.com account, which allows one concurrent build at a time and unlimited time and memory resources.


drmeister:
Travis is not going to work for us

PR #447, opened 2018-03-17
Fix logical-pathname & (pathname nil)

kpoeck:
Logical-Pathname was not correctly exported to lisp. Additionally the error handling threw an exception.

Test with 
````
COMMON-LISP-USER> (logical-pathname "MODULES:nada;foo;bar.lisp")            

#P"MODULES:NADA;FOO;BAR.LISP"
````
Error case:
````
COMMON-LISP-USER> (logical-pathname "/Users/karstenpoeck/lisp/ansi-cl-tests/nada.lisp")
Condition of type: SIMPLE-TYPE-ERROR
#P"/Users/karstenpoeck/lisp/ansi-cl-tests/nada.lisp" cannot be coerced to a logical pathname.
````

Additionally (pathname nil) didn't throw a type-error, does now

### DIFF
--- a/src/core/pathname.cc
+++ b/src/core/pathname.cc
@@ -979,7 +979,9 @@ struct PathnameRecursionGuard {
 CL_DEFUN Pathname_sp cl__pathname(T_sp x) {
   PathnameRecursionGuard guard;
   if (x.nilp()) {
-    SIMPLE_ERROR(BF("The only argument for pathname is nil"));
+    // knpk also needs to be a type error
+    // SIMPLE_ERROR(BF("The only argument for pathname is nil"));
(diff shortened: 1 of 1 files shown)


Bike:
@kpoeck, i'm going to merge these, but I want to note a few style things-

Please don't put in a comment with just your name. Your name is properly associated with the change in git's history.

If you're fixing something small or obvious like this, you don't need to keep the old code in a comment, or describe what errors it gave you. Commenting out old code only makes sense if there's any chance need it again, or if you need to explain a complex way to do it that doesn't work, to head off future maintainers.

Thanks for the help, just wanna keep the codebase from getting any messier than it already is

Issue #450, opened 2018-03-18
Ansi-testst: Read-sequence error


kpoeck:
````
(WITH-INPUT-FROM-STRING (stream "sadbasd")
	   (READ-SEQUENCE "1234" stream))
->
#<TYPE-ERROR> [#<STRING-INPUT-STREAM > is not of type CORE:STRING-OUTPUT-STREAM.]
````

kpoeck:
Have a fix, will submit PR

kpoeck:
Fixed by PR #452

Bike:
This example isn't valid, is it? "1234" is a string literal, you can't modify it.

kpoeck:
Bike is right, so better example:
````
(let ((string (make-string 4 :initial-element #\A)))
	   (WITH-INPUT-FROM-STRING (stream "sadbasd")
	     (READ-SEQUENCE string stream)
	     string))
"sadb"
````

Issue #449, opened 2018-03-18
Ansi-tests: (stream-element-type  (MAKE-CONCATENATED-STREAM *STANDARD-INPUT*)) signals error


kpoeck:
````
COMMON-LISP-USER> (stream-element-type  (MAKE-CONCATENATED-STREAM *STANDARD-INPUT*))

Condition of type: TYPE-ERROR
#<CONCATENATED-STREAM > is not of type BROADCAST-STREAM.

Available restarts:
(use :r1 to invoke restart 1)

1. (RESTART-TOPLEVEL) Go back to Top-Level REPL.
````

kpoeck:
I have a fix, will submit a PR

kpoeck:
pr #452

kpoeck:
works, but unfortunately I seem to have made the fix against master, not dev.
Please advise how to continue

Bike:
we'll fix it up later

kpoeck:
Was merged in the meantime, closing

PR #452, opened 2018-03-18
Fix stream-element-type of MAKE-CONCATENATED-STREAM, str_in_element_type

kpoeck:
Fixes error:
#450 (probaby copy-paste error)
#449  Concatenated streams used the element-type of boradcast streams


kpoeck:
@Bike is it possible that you merged this into master and not into dev?

kpoeck:
Possibly  because i did the pr to thebwrong branch. Please advise how to fix this

Issue #470, opened 2018-03-29
Ansi-test: make-symbols should accept all strings, not only simple-strings


kpoeck:
````
COMMON-LISP-USER> (make-symbol (MAKE-ARRAY '(6)
					  :INITIAL-CONTENTS '(#\A #\B #\C #\D #\E #\F)
					  :ELEMENT-TYPE 'CHARACTER
					:fill-pointer 4))

Condition of type: TYPE-ERROR
"ABCD" is not of type SIMPLE-STRING.

Available restarts:
(use :r1 to invoke restart 1)
````

Bike:
dupe of #370

kpoeck:
Could you verify pr #471 for both issues

kpoeck:
Fixed

PR #471, opened 2018-03-29
Fix make-symbol to accepts all kind of strings , not only simple-strings

kpoeck:
Fixes issue #470.

To test:
````
(defun make-symbol-tests ()
  (dolist (test (list 23 'defun (code-char 65) ))
    (assert
     (typep 
      (handler-case
	  (make-symbol test)
	(error (e) e))
      'type-error)))
  (dolist (test (list "ABCCC" (MAKE-ARRAY '(6)
					  :INITIAL-CONTENTS '(#\A #\B #\C #\D #\E #\F)
					  :ELEMENT-TYPE 'CHARACTER
					  :FILL-POINTER 4)))
    (assert
     (symbolp 
      (handler-case
	  (make-symbol test)
	(error (e) e))))
    )
  )
````

Will try to find the reference from bike regarding regression-testing and add the tests in appropriate form there

### DIFF
--- a/src/core/symbol.cc
+++ b/src/core/symbol.cc
@@ -38,6 +38,7 @@ THE SOFTWARE.
 #include <clasp/core/package.h>
 #include <clasp/core/lisp.h>
 #include <clasp/core/wrappers.h>
+#include <clasp/core/designators.h>
 
 // Print more information about the symbol
 #define VERBOSE_SYMBOLS 0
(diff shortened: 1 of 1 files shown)


Bike:
Tests go in src/lisp/regression-tests. it's pretty bare but you can get the idea from the other files in the directory. probably we'd want a new symbols.lisp file.

PR #476, opened 2018-04-01
Added minimal extensions directory structure with wscript adaptions.

dg1sbg:
Issue:
When adding multiple extensions then questions is how to include all those in the build process. Also, there was an include directive in src/gctools/gc_interface.cc saying "#include <project_headers.h>" - which only worked for cando and not for multiple extensions.
Solution Idea:
Have a well-defined directory structure below clasp/extensions where:
1. Each extension has its own sub-directory, e.g. "dummy"
2. Have an include directory directly below clasp/extensions with one include file: extensions.h
3. In this include file all extensions' toplevel include file are inserted manually when adding an extension, e.g. "#include <dummy/dummy.h>" for extension dummy.
4. Each extension directory has at least sub-directories "src" and "include". The wscript file in the extensions root directory has to recurse down into these directories.
5. Also, the wscript file in the extension's root directory has to load a wscript.config file in which a config flag "CLASP_EXT_ENABLE_DUMMY = 1" enables the build of that extension. This variable has to be checked in the extensions' root directory wscript file, and, if set to 0, will effect to not configure/build the extension.
All this assumes we distribute the wscript files in the extensions' dirs. @attila-lendvai Do you have another idea for this? @drmeister I'd need to have your thoughts on this as I am heading into several extensions at once ... 
Thank you all for providing feedback!

### DIFF
--- a/extensions/include/extensions.h
+++ b/extensions/include/extensions.h
@@ -1,0 +1,42 @@
+/*
+    File: extensions.h
+*/
+
+/*
+Copyright (c) 2018, Christian E. Schafmeister
+
+CLASP is free software; you can redistribute it and/or
+modify it under the terms of the GNU Library General Public
+License as published by the Free Software Foundation; either
+version 2 of the License, or (at your option) any later version.
+
+See directory 'clasp/licenses' for full details.
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+/* -^- */
+
+
+#ifndef CLASP_EXTENSIONS_H //[
+#define CLASP_EXTENSIONS_H
+
+/* THIS FILE IS EMPTY ON PURPOSE AS LONG AS NO EXTENSIONS ARE IMPLELEMTED */
+/* ANY EXTENSION MUST PLACE A TOPLEVEL INCLUDE FILE HERE FOR GC INCLUSION */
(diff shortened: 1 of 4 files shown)


attila-lendvai:
i don't yet have a strong opinion on extensions. with that in mind, in general i would introduce a custom python API defined by the toplevel wscript and used by the wscript files under extensions/ (e.g. to register extra files for the build). i have already set up the infrastructure for that, logging is an example (see in cando's wscript).

i have created a section on the wiki to collect the requirements: https://github.com/clasp-developers/clasp/wiki/The-Build-Process

@dg1sbg please take a look and extend if something is missing.

the rationale is that the extension's build script will need to mess with the master build script's internals anyway, e.g. to register files, or to compile stuff only in specific stages, etc... so, if they need to know about the internals of the master wscript, then they might as well call functions with fancy names, hiding some of the internals, and communicating some of the intentions.

meanwhile, i think the need for 5) has somewhat been obsoleted by one of drmeister's recent patches that introduced the dboehm target. you can think of it this way: up until the 'c' stage (=4) you are building vanilla clasp without extensions, and in stage 'd' (=5) you're building an app that is clasp + extensions/.

drmeister:
I need to sleep on this one - and try it out on a separate clone.

drmeister:
I haven't moved on the extensions pull request because I had a bunch of demos of the Cando extension and I didn't want to cause myself problems.  We can move forward now.
Does that pull request still capture your latest thinking?
Because there is one thing that it doesn't capture - the worst problem of all.  The clasp_gc.cc file that the static analyzer generates needs to be generated for every combination and permutation of extensions.
You can build the boehm version of clasp without clasp_gc.cc   but the MPS version needs have a clasp_gc.cc file that was calculated from a specific combination of clasp+whatever extensions.   
I have no ideas about how to do it some other way.
My best idea is to incorporate the static analyzer into the build - that would mean building the boehm version, running the static analyzer in the boehm version on the C++ source code, building the MPS version.   That would add a few hours to the MPS build.
The second idea is to minimize the number of extensions to things that only add classes that the GC needs to manage or to differentiate between extensions that need the static analyzer and those that don't.
For your extensions - do you add classes that the GC needs to manage?

dg1sbg:
Well, there's a couple points I'd like to explore / have feedback on - I had seen and seem to have had understood the clasp_gc.cc mechanism. I wanted to have a way of "simple include" solution by introducing an extensions/include/gc.h file where developers need to put into the include statements that included the respective package's gc file.
But:
1. I would imagine a way of "registering" classes that the static analyzer needs to be aware of.
2. I would like to avoid having to manually change a clasp-internal include file or even source file.

And yes, I, too, have now real good solution to this. As we are at the C++ level we might have a solution where we have no global clasp_gc.cc file anymore but have the static analyzer search for XXXX_gc.cc files and read those ... So effectively spreading the gc relevant declarations into normal .cc files wherever they seem to be fit in terms of directory structure. The static analyzer or another tool would then gather the respective XXXXX_gc,h files and patch the clasp_gc.cc file automatically. This would be more safe as we don't have a manual step involving several developers... Am I making sense?

KR Frank

PR #477, opened 2018-04-01
Add regression tests for my recent Prs concerning the ansi-tests

kpoeck:
Added test according to the framework in regression-tests that bike explained.

Tested the test both in a recent clasp, clasp with all my changes, and -if possible - in sbcl


drmeister:
Merged - thank you!

PR #478, opened 2018-04-01
Allow defstruct with keywords as names

### LINKED ISSUE #475: Ansi-tests: (defstruct :otto a b c) and (defstruct (:otto (:type vector)) a b c) and (defstruct (:otto (:type list)) a b c) fail
````
(defstruct (:otto (:type list)) a b c)
->
In this type of lambda list, the required parameter must
be a variable which is also not the name of a constant.
But the following was found instead:
:OTTO
   [Condition of type CLEAVIR-CODE-UTILITIES:REQUIRED-MUST-BE-VARIABLE]
````

kpoeck:
Fixes #475 
tests added as possible in previous pr


Bike:
going to assume the reverts are fine and normal

PR #479, opened 2018-04-03
Signal a program-error if charxx functions are called with no arguments

kpoeck:
See tests in pr #477 , src/lisp/regression-tests/character0.lisp.

Example:
````
COMMON-LISP-USER> (handler-case (char=) (program-error (e) e))
#<PROGRAM-ERROR>
````


drmeister:
Got it - thank you!

PR #481, opened 2018-04-05
Correctly treat code-chars > 255

kpoeck:
Wrong:
````
COMMON-LISP-USER>> (let ((*print-escape*)) (write (code-char 256)))
#\NUL
````
Right:
````
COMMON-LISP-USER> (let ((*print-escape*)) (write (code-char 256)))
Ā
#\U100
````


drmeister:
Merged

PR #485, opened 2018-04-06
Feature nil bitarrays

kpoeck:
Allow (APPLY 'BIT-AND '(#0A0 #0A0)) and friends.

Se test-suite in bit-array0.lisp of pr #477


drmeister:
Do I read this correctly?  Is handle-empty-bitarrray a runtime check on the binary operator?
If so and since we know the binary operator at compile time - can we reduce it to only the needed check at runtime?   
I may be reading the code incorrectly.

kpoeck:
The Problem I am trying to solve is the following:
a Bitarray can be 1 one of the following cases:

- (SIMPLE-BIT-VECTOR *) as in (type-of (make-array 2 :element-type 'bit :initial-element 0))
- (SIMPLE-ARRAY BIT <dimensions> ) as in (type-of (make-array (list 2 2) :element-type 'bit :initial-element 0)) or (type-of (make-array nil :element-type 'bit :initial-element 0))
- (ARRAY BIT (2 2)) as in (type-of (make-array (list 2 2) :element-type 'bit :initial-element 0 :adjustable t))

CL_DEFUN T_sp core__bit_array_op(int opval, T_sp tx, T_sp ty, T_sp tr) .. only handles correctly the SIMPLE-BIT-VECTOR case and fails for simple-array and array

kpoeck:
The first problem ist that asAbstractSimpleVectorRange(AbstractSimpleVector_sp for SimpleMDArrayBit_O uses length() that is for a mdarray is defined as 0xDEADBEEF01234567 and than all hell breaks loose.

This I already fixed. (Pr pedning but code is:)
````
void asAbstractSimpleVectorRange(AbstractSimpleVector_sp& sv, size_t& start, size_t& end) const final {
      sv = gc::As<SimpleBitVector_sp>(this->_Data);
      start = this->_DisplacedIndexOffset;
      // this->length() is the dummy value in this case, 0xDEADBEEF01234567
      // this->length() in this class shoud return an error
      // end = this->length()+this->_DisplacedIndexOffset;
      end = sv->length()+this->_DisplacedIndexOffset;
    }
````

kpoeck:
the second problem is that T_sp core__bit_array_op always returns a SIMPLE-BIT-VECTOR, even if the inout was a simple-array or even an error.

Am currently trying to figure out how to handle this.

The lisp-code in this pr just handles the case of nil bitarrays as in (make-array nil :element-type 'bit :initial-element 0), I now wonder whether it is the best to handle this in lisp or in c++

kpoeck:
I have the asAbstractSimpleVectorRange, pr tomorrow. Right now looking at core__bit_array_op returning the right class, not always a simpleVector

kpoeck:
Hello drmeister,
finally I send the c++ fixes separately in pr#500.

The fixes here just fix the case where operands of bit-* have nil dimensions as in (make-array nil :element-type 'bit :initial-element 0).

The cost is a a runtime-check for the operand.

Additionally - but that was like this before - the operator is passed to c++. If you want to avoid that, thats easy enough, than we have to call core__bit_array_op_b_clr_op and friends, but that calls again core__bit_array_op with opval as the firt parameter

drmeister:
Excellent thank you.

PR #486, opened 2018-04-06
clos: don't allow subclassing builtin classes

dkochmanski:
This issue was discussed at some point in the past. See
https://gitlab.com/embeddable-common-lisp/ecl/issues/426 for some logs.


dkochmanski:
new commit solves problem which was previously masked by having `*clos-booted*` not being `T` (namely forward-referenced classes support). See commit message for details.

drmeister:
Merged!  Thank you very much!

PR #488, opened 2018-04-09
<= and friends should throw a program-error when called with no argument

kpoeck:
now as specified:
````
COMMON-LISP-USER> (<=)

Condition of type: PROGRAM-ERROR
#<PROGRAM-ERROR>
````


drmeister:
Merged - thank you!

PR #491, opened 2018-04-14
Update makefile to changes in wscript

kpoeck:
use update_dependencies instead of update_submodules


kpoeck:
I think after the PR Attila changed the file in the same way

attila-lendvai:
thanks! i pushed a separate commit though, this PR contains a lot of extra stuff...

PR #498, opened 2018-05-01
Feature last take 2

kpoeck:
Fixes a lot of edge-cases (negative indices, most-positive-fixnum, bignums, ...) for cl__nth, cl__nthcdr, cl__last, cl__nbutlast, cl__butlast.

Removes some useless comments that I introduces before

See sequences01.lisp for regression-tests for the above


Bike:
With regards to the `TYPE_ERROR`s, to have a cons as a type you have to make one. For example, here is a type error for the type `(or string null)`: https://github.com/clasp-developers/clasp/blob/master/src/core/package.cc#L282

But I think `unsigned-byte` is fine so I'll merge this.

PR #503, opened 2018-05-11
Fix oddp for negative numbers + tests

kpoeck:
-1 % 2 returns -1 and not 1 as in the previous code.
Added test for all edge cases


drmeister:
Thank you!

PR #505, opened 2018-05-13
Working port to FreeBSD.  Tested on -11.

cracauer:
Caveats:
- no wscript config (have elsewhere)
- no FreeBSD port (as in for /usr/ports), which I also have elsewhere


drmeister:
Thank you very much!

PR #508, opened 2018-05-17
Feature characters more unicode

kpoeck:
more clasp characters more unicode compliant
see regression-tests for tests
passes all ansi-tests for characters


drmeister:
Merged!  Thank you!

PR #510, opened 2018-05-19
Fix more ansi exceptions

### LINKED ISSUE #366: Intrinsics signal the wrong types of error
`((lambda () x))` signals a CORE:SIMPLE-PROGRAM-ERROR. It ought to signal CL:UNBOUND-VARIABLE.

Similar with unbound functions and maybe a few other things. The error signaling code is intrinsic_error in src/llvmo/link_intrinsics.cc. Probably they just need to use UNBOUND_VARIABLE_ERROR and stuff.

kpoeck:
Should also fix #366 
drops useless first parameter for core__sequence_start_end


drmeister:
Thank you very much!

PR #511, opened 2018-05-19
Some fixes to make packages safer

kpoeck:
Mark package :cl, :core and :keyword as system-locked
Don't unimport in these packages, also no symbols with home-package being locked
don't delete system-locked packages (added core package to test)
This is just a start in the implementation of locked-packages as in sbcl

delete nicknames in delete-packages, so that they can be used again
See the regression-tests in package.lisp


drmeister:
Package safety!  I like it - thank you very much!

PR #512, opened 2018-05-19
Some string fixes

kpoeck:
Return correct indices for the string functions


src/core/array.cc
```diff
@@ -2057,6 +2059,7 @@ Array_sp SimpleBitVector_O::unsafe_setf_subseq(size_t start, size_t end, Array_s
   TYPE_ERROR(other,cl::_sym_bit_vector);
 }
 
+// This does not work properly
```
drmeister: Does it need to be removed or fixed?

kpoeck:
Needs to be fixed, see issue 436

kpoeck:
The rest of the fixes should be fine though

kpoeck:
@drmeister , have now fixed Fix SimpleBitVector_O::unsafe_fillArrayWithElt, so all should be fine now. 
Tested with more regression-tests and ansi-tests

drmeister:
Merged - thank you!

PR #514, opened 2018-05-29
Fix issue 456

### LINKED ISSUE #456: Ansi-tests:  (format t "~3,f" 1.0) ->
````
COMMON-LISP-USER>  (format t "~3,f" 1.0)

Condition of type: TYPE-ERROR
0 is not of type REAL.

Available restarts:
(use :r1 to invoke restart 1)
````

kpoeck:
````
COMMON-LISP-USER> (format t "~3,f" 1.0)
1.0
````
fixes #456 

There is something fishy in the glue code, this just failed for "0" as the argument to core__float_to_digits position


drmeister:
Merged - thank you.

PR #518, opened 2018-06-05
Programmatically enable HANDLE_FORK in boehm_gc at runtime

d235j:
This should remove the requirement of having to rebuild the Boehm garbage collector with HANDLE_FORK.

Most builds of the garbage collector have HANDLE_FORK support compiled in, just disabled at runtime. This behavior will change in version 7.7.0 of the garbage collector, but we can programmatically enable it by calling GC_set_handle_fork() before GC_INIT().
If a version of the garbage collector does not have it compiled in, and the platform supports threading, the garbage collector will simply abort (see the logic at https://github.com/ivmai/bdwgc/blob/cb1194d/misc.c#L214).


drmeister:
Merged!  Thank you!

PR #528, opened 2018-06-13
Fix waf nodes for lisp files

kpoeck:
assert used the wrong variable


attila-lendvai:
thanks for the fix!

do you know why you have all those "Merge upstream" commits in this PR? can you eliminate them?

attila-lendvai:
i've cherry picked and pushed it manually. thanks!

attila-lendvai:
@kpoeck, instead of

`
git merge --no-edit -m "Merge Upstream" upstream/dev
`

you can do:

`
git rebase -v upstream/dev
`

or if you don't record any commits of your own into your local `dev` branch, i.e. you just want to follow what's in upstream, then simply:

`
git pull upstream
`

your local `dev` branch is now probably full of those merge commits. you need to get rid of them. look at the output of `git log`, find a hash that is old enough to be prior to the first local merge commit, then `git reset --hard [hash]` to rewind to that commit, and then `git pull upstream`.

watch out not to delete any of your commits that you want to keep!

kpoeck:
I don’t have any own commits in dev, so that should work

PR #530, opened 2018-06-15
MPS fork support

dlovemore:
We now support fork() in MPS and parallel builds are working in clasp.

Please read:

https://www.ravenbrook.com/project/mps/master/manual/html/topic/thread.html#fork-safety

for information and caveats to check if there might be further work to do clasp side for safety and robustness.


drmeister:
Thank you

PR #533, opened 2018-06-24
Feature print float infinity

kpoeck:
Fixes
````
COMMON-LISP-USER> (let ((*read-default-float-format*  'single-float)
      (*print-readably* nil))
  (read-from-string (format nil "12~40,2f" most-positive-single-float)))

#.core::single-float-positive-infinity
````


Bike:
Given that there's no `trap-fpe`, we can probably lose the unwind-protect too

But more importantly, we should export the float infinities from ext rather than leave them in core.

kpoeck:
can change to ext:
what is the proper place to export from ext?

kpoeck:
@Bike, hope to have done it now properly, please review

PR #534, opened 2018-06-24
Fix float reader ofr short-float & long-float

kpoeck:
Before:
````
COMMON-LISP-USER> (equalp (list 1.111 1.111 1.111d0 1.111d0)
              (let ((result nil))
                (dolist (type (list 'short-float 'single-float 'double-float 'long-float) (reverse result))
                  (let ((*read-default-float-format* type))
                    (push (read-from-string "1.111") result)))))

Condition of type: SIMPLE-PROGRAM-ERROR
Handle *read-default-float-format* of SHORT-FLOAT

Available restarts:
(use :r1 to invoke restart 1)

````

After
````
COMMON-LISP-USER> (equalp (list 1.111 1.111 1.111d0 1.111d0)
              (let ((result nil))
                (dolist (type (list 'short-float 'single-float 'double-float 'long-float) (reverse result))
                  (let ((*read-default-float-format* type))
                    (push (read-from-string "1.111") result)))))

T
````

### DIFF
--- a/src/core/lispReader.cc
+++ b/src/core/lispReader.cc
@@ -856,13 +856,26 @@ T_sp interpret_token_or_throw_reader_error(T_sp sin, Token &token, bool only_dot
         char *lastValid = NULL;
         if (cl::_sym_STARreadDefaultFloatFormatSTAR->symbolValue() == cl::_sym_single_float) {
           string numstr = tokenStr(sin,token, start - token.data())->get_std_string();
+          // don't understand this, clasp_make_single_float takes a float , not a double
+          // shouldn't that be float f = strtof(......
           double d = ::strtod(numstr.c_str(), &lastValid);
           return clasp_make_single_float(d);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/core/lispReader.cc [hunk 1]
```diff
@@ -856,13 +856,26 @@ T_sp interpret_token_or_throw_reader_error(T_sp sin, Token &token, bool only_dot
         char *lastValid = NULL;
         if (cl::_sym_STARreadDefaultFloatFormatSTAR->symbolValue() == cl::_sym_single_float) {
           string numstr = tokenStr(sin,token, start - token.data())->get_std_string();
+          // don't understand this, clasp_make_single_float takes a float , not a double
+          // shouldn't that be float f = strtof(......
           double d = ::strtod(numstr.c_str(), &lastValid);
```
drmeister: I agree - if you would like to make the change to use strtof - go ahead -

drmeister:
Thank you very much!

PR #548, opened 2018-07-17
throw proper error if file passed to compile-file does not exist

kpoeck:
one more for the ansi-tests


kpoeck:
I forgot tho add :pathname to the  (error 'core:simple-file-error ..., will update in a moment

drmeister:
Thank you very much kpoeck!

PR #550, opened 2018-07-22
Fix #292 by adding much more type-tests

### LINKED ISSUE #292: RANDOM accepts negative bounds
(random 0.0) or (random -1) crashes somewhere in boost.

kpoeck:
See included regression-test and #292 for what is being fixed


drmeister:
Thank you.

PR #554, opened 2018-07-22
Better package exceptions, cleanup some exceptions

kpoeck:
See tests fore cases fixed and tested
Synched exceptions.h and conditions.lisp
Missing core:closed-stream that is defined in exceptions.h and not in conditions.lisp (different pr)


drmeister:
Better error messages - thank you!

PR #557, opened 2018-07-26
Use single shared chain between AMC pools.

dlovemore:
This is more efficient. See:

https://www.ravenbrook.com/project/mps/master/manual/html/topic/collection.html#generation-chains


drmeister:
Thank you very much - I added the cons pool waaaaay after I first added the general pool.  I missed the point about two pools sharing the same chain.

PR #564, opened 2018-07-29
Fix pathname* to throw a type-error, when nil is passed as stream, instead of simple-error

kpoeck:
e.g.
```lisp
COMMON-LISP-USER> (pathname nil)

Condition of type: SIMPLE-TYPE-ERROR
In function PATHNAME,the value of the only argument is
  NIL
which is not of the expected type (OR FILE-STREAM STRING PATHNAME SYNONYM-STREAM)

Available restarts:
(use :r1 to invoke restart 1)

````
validated running the ansi-tests

### DIFF
--- a/src/core/pathname.cc
+++ b/src/core/pathname.cc
@@ -711,7 +711,8 @@ CL_DEFUN T_sp core__pathname_translations(T_sp host, T_sp hostp, T_sp set) {
         T_sp item = CAR(l);
         T_sp from = coerce_to_from_pathname(oCar(item), host);
         T_sp pn = oCadr(item);
-        if (pn.nilp()) SIMPLE_ERROR(BF("%s was about to pass nil to pathname") % __FUNCTION__);
+        if (pn.nilp())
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/core/pathname.cc [hunk 1]
```diff
@@ -988,11 +989,12 @@ CL_DEFUN Pathname_sp cl__pathname(T_sp x) {
     x = cl__parse_namestring(x);
   } else if (gc::IsA<Pathname_sp>(x)) {
     // do nothing
-  } else if (gc::IsA<Stream_sp>(x)) {
+  } else if ((gc::IsA<FileStream_sp>(x)) or (gc::IsA<SynonymStream_sp>(x))){
```
drmeister: I think this will be the first 'or' in the Clasp code. 
Should we allow this?  Or stick with || ???

kpoeck:
Will change (I am happy to follow any code conventions as long as I know them)

kpoeck:
@drmeiter Now have       ((gc::IsA<FileStream_sp>(x)) || (gc::IsA<SynonymStream_sp>(x))), please check

drmeister:
Thanks!

PR #566, opened 2018-07-29
Various fixes in for inline

kpoeck:
- elt for lists cannot be transferred to nth, since nth does not error, when the index is too large, use nthcdr instead
- elt for vectors similar problem, cannot be transferred symply to aref, since fill-pointer treament is different (elt need to respect, aref not)
- more precise errors, if index out of bounds
- treat empty arrays

### DIFF
--- a/src/lisp/kernel/cleavir/inline.lisp
+++ b/src/lisp/kernel/cleavir/inline.lisp
@@ -538,6 +538,20 @@
             (ext:byte16 nil) (ext:byte8 nil)
             (bit t))))
 
+;;; If the array has a fill-pointer, this might be wrong, at least when called from elt
+;;; (elt (make-array 10 :initial-contents '(0 1 2 3 4 5 6 7 8 9) :fill-pointer 3) 5)
+;;; sbcl -> The index 5 is too large, but no error in clasp
+;;; aref allows access further than the fill pointer
+(declaim (inline vector-in-bounds-p))
+(defun vector-in-bounds-p (vector index)
+  (etypecase vector
+    ((simple-vector *)
+     (and (<= 0 index) (< index (core::vector-length vector))))
+    (array
+     (and (<= 0 index) (< index (if (array-has-fill-pointer-p vector)
+                                    (min (core::%array-total-size vector) (fill-pointer vector))
+                                    (core::%array-total-size vector)))))))
+
 (declaim (inline row-major-array-in-bounds-p))
 (defun row-major-array-in-bounds-p (array index)
   (etypecase array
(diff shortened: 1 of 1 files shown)


kpoeck:
@Bike , since this is your territory, please review

drmeister:
Thanks kpoeck - I'll wait until Bike takes a look at this.

Bike:
non-empty-array is an unnecessary complication. Better signal a simple-type-error explaining that accessing a length zero vector is dumb.

fast-vector-access isn't a good name for "access a vector but respect fill pointer" or "row-major-aref, but only from elt". It's doing kind of a lot of type checks- first for with-array-data, then in vector-in-bounds-p. it might be worth having with-array-data return a maximum while it does its thing.

If elt is broken setf-elt is probably also broken

kpoeck:
I am not sure how to avoid the non-empty-array.
Assume I do something like:
```lisp
(error 'simple-type-error :FORMAT-CONTROL "Array not valid ~a for index ~a ~%" :format-arguments (list vector index))
````
That prints fine, but does not use the :datum and :expected-type keywords with might be not-ok for a type-error. Actually all uses of simple-type-error use all 4 keywords (exception adjust-array, that uses 4, probably a bug)
if I do
```lisp
(error 'simple-type-error :datum vector :expected-type 'array :FORMAT-CONTROL "Array not valid ~a for index ~a ~%" :format-arguments (list vector index))
````
than I am not telling the truth, since vector is an array, just with no elements and I am where I was before

Thinking more about it, I don't really like that  simple-type-error inherits from simple-condition and thereby overwrites the :REPORT of type-error. Probably be better that type-error already inherits from condition or simple condition. Same confusion with core:simple-file-error overwriting the :report of  file-error. 

Please advise

kpoeck:
I think I have worked on your concerns:
- non-empty-array is gone, simple-type-error used
- fast-vector-access renamed
- setf-elt  also fixed
- Test cases:
```lisp
(test-expect-error elt-1 
		   (ELT "ABCDEFGH" -1)
		   :type type-error)

(test-expect-error elt-2 
		   (ELT "ABCDEFGH" 15)
		   :type type-error)

(test-expect-error elt-3 
		   (ELT (make-array 10 :initial-element 42) 15)
		   :type type-error)

(test-expect-error elt-4 
		   (ELT (make-array 10 :initial-element 42) -1)
		   :type type-error)

(test-expect-error elt-5 
		   (ELT (list 1 2 3) -1)
		   :type type-error)

;;; 
(test-expect-error elt-5a 
		   (ELT (list 1 2 3) 4)
		   :type type-error)

(test-expect-error elt-5b 
		   (ELT (cons 1 2) 1)
		   :type type-error)

(test-expect-error elt-6 
		   (ELT nil 2)
		   :type type-error)

(test-expect-error elt-7 
		   (ELT (make-array 0) 0)
		   :type type-error)

;;; fails, should report an error, since 5 is after the fill-pointer
;;; for aref should not fail, since elt is convert to aref, we have a deviation
;;; fixed now
(test-expect-error elt-8 
		   (elt (make-array 10 :initial-contents '(0 1 2 3 4 5 6 7 8 9) :fill-pointer 3) 5)
		   :type type-error)

(test-expect-error elt-9 
		   (setf (ELT (list 1 2 3) -1) 23)
		   :type type-error)

(test-expect-error elt-10 
		   (setf (ELT (list 1 2 3) 4) 23)
		   :type type-error)

(test-expect-error elt-11 
		   (setf (ELT nil 2) 23)
		   :type type-error)

(test-expect-error elt-12 
		   (setf (ELT (make-array 0) 0) 23)
		   :type type-error)

(test-expect-error elt-13 
		   (setf (elt (make-array 10 :initial-contents '(0 1 2 3 4 5 6 7 8 9) :fill-pointer 3) 5) 23)
		   :type type-error)
````

PR #572, opened 2018-07-29
Batch up files in parallel build

dlovemore:
Parallel builds batch up to 10 files at a time to reduce fork() overhead.

This is a pragmatic step to save about 5-7 minutes of build time.


src/lisp/kernel/clasp-builder.lsp
```diff
@@ -343,19 +377,19 @@ Return files."
...
+                   (when reload (reload-some finished-entries))
                    (decf child-count))
                  (error "wait returned ~d  status ~d~%" wpid status)))
-           (when entry
+           (when entries
              (multiple-value-bind (maybe-error pid-or-error result-stream)
                  (core:fork t)
                (if maybe-error
-                   (error "Could not fork when trying to build ~a" entry)
+                   (error "Could not fork when trying to build ~a" entries)
```
dlovemore: Is this correct? I am not sure what ~a does. entries is a list.

drmeister: Yes - it's correct. The ~a means "print argument in an Aesthetically pleasing way".

PR #573, opened 2018-07-31
Move reader-error to package core

kpoeck:
Since according to clhs is not part of (find-package :cl)
Also synch the parameters in  CL_LAMBDA and effective function definition


Bike:
`reader-error` is actually a defined CL symbol, for the class- but per 11.1.2.1.1, we can't have it name something else, which is the real basis for this fix.

PR #580, opened 2018-08-21
Fix asoc for nil in list, correct type-error

kpoeck:
Fix c++ side of 
```lisp
(ASSOC NIL '((A . B) NIL (C . D) (NIL . E) (NIL . F) NIL (G . H)))
-> 
(NIL . E)
````
Fix c++ side of error message for
```lisp
(assoc 3 (list (list 1 3) 2))
-> 
Condition of type: TYPE-ERROR
2 is not of type CONS.
````
Tests  - separate pr -
```lisp
(test assoc-1
      ;;; avoid the compiler-macro
      (eval '(equal (cons nil 'e)(ASSOC NIL '((A . B) NIL (C . D) (NIL . E) (NIL . F) NIL (G . H))))))

(test assoc-compiler-macro
      (equal (cons nil 'e) (Let () (ASSOC NIL '((A . B) NIL (C . D) (NIL . E) (NIL . F) NIL (G . H))))))

(test-expect-error assoc-2-error
                   (assoc 3 (list (list 1 3) 2))
                   :type type-error)

(test-expect-error assoc-2-error-no-compiler-macro
                   (eval '(assoc 3 (list (list 1 3) 2)))
                   :type type-error)
````

### DIFF
--- a/src/core/cons.cc
+++ b/src/core/cons.cc
@@ -308,12 +308,16 @@ List_sp Cons_O::assoc(T_sp item, T_sp key, T_sp test, T_sp testNot) const {
   Tester t(item, key, test, testNot, false);
   for (auto cur : (List_sp) this->asSmartPtr()) {
     LOG(BF("Testing for assoc with item=%s entry = %s") % item % oCar(cur));
-    if (CONS_CAR(cur).consp()) {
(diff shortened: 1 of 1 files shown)


Bike:
This is nitpicky, but the type in the error should be `list`, since after all `nil` is allowed.

PR #598, opened 2018-09-10
Fix crash of (MAKE-sequence 'list -1), add regression tests

kpoeck:
- Fixes (MAKE-sequence 'vector -1) to not crash
- Fixes (MAKE-sequence 'cons 0) to error
also contains regression tests for other sequence errors


Bike:
- indentation is messed up.
- you have the `make-sequence` expansion doing `subtypep` at runtime even though both types are known at compile time. If the type is not a subtype of cons - which is probably the usual case anyway - it shouldn't even check `zerop`.
- The comment in the actual `make-sequence` function isn't really helpful - why not test for null too?
- A better way to do this might be to rewrite `closest-sequence-type` to return a minimum length in addition to the exact length.

PR #604, opened 2018-09-13
Verify to displaced-to array is big enough 

kpoeck:
for the displaced-index-offset
```lisp
(test-expect-error make-array-6
                   (make-array 5 :element-type (array-element-type "") :displaced-index-offset 2 :displaced-to "")
                   :type simple-error)
````

### DIFF
--- a/src/lisp/kernel/lsp/arraylib.lsp
+++ b/src/lisp/kernel/lsp/arraylib.lsp
@@ -75,7 +75,9 @@ contiguous block."
     (if (let ((array-element-type (array-element-type displaced-to)))
           (not (and (subtypep array-element-type upgraded-element-type)
                     (subtypep upgraded-element-type array-element-type))))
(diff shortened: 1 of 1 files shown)


Bike:
`core::%array-total-size` is a special operator in cclasp. Before that it won't work, unless I'm forgetting something.

kpoeck:
Are you saying, won't work in aclasp and bclasp? If so, is there any alternative?

Bike:
That's what I mean, yes. You can probably just use `cl:array-total-size` - it won't be as fast as possible, but we're allocating a displaced array anyway. Also in an ideal world it would end up with the special operator if possible without you specifically asking for it.

While you're at it, the error message should probably message the displaced array index and the array size that's too small for it.

kpoeck:
I copied the message literally from ecl :-), but obviously your proposal is better

kpoeck:
@Bike: Your comments should be included now

Issue #579, opened 2018-08-15
(- most-negative-fixnum) -> -2305843009213693952


kpoeck:
should be 2305843009213693952

kpoeck:
there is a compile macro calling core:negate, but is not obvious what is called later

Bike:
"later"?

Anyway, `(core:negate most-negative-fixnum)` returns this as well, as does an uncompilermacroed `(- most-negative-fixnum)`. The former is the important one, because the `-` function will call `negate` if there are no more arguments.

`clasp_negate` tries to return a fixnum if given a fixnum, but that doesn't work for `most-negative-fixnum`. That's probably the issue.

PR #606, opened 2018-09-15
Fix printing  negative integers at base 16

### LINKED ISSUE #605: ansi-tests: printing of negative numbers in hex (and possible other bases) broken
```lisp
"-1"
CL-USER> (LET ((*PRINT-BASE* 16))
             (write -1))
FFFFFFFFFFFFFFFF
-1
````

kpoeck:
Fixes #605 
Test:
```lisp
(test print-1
      (string=
       "-1"
       (LET ((*PRINT-BASE* 16))
         (WITH-OUTPUT-TO-STRING (*STANDARD-OUTPUT*) (write -1)))))
````

### DIFF
--- a/src/core/numberToString.cc
+++ b/src/core/numberToString.cc
@@ -109,6 +109,10 @@ CL_DEFUN StrNs_sp core__integer_to_string(StrNs_sp buffer, Integer_sp integer,
   if (integer.fixnump()) {
     char txt[64];
     gc::Fixnum fn = unbox_fixnum(gc::As<Fixnum_sp>(integer));
+    if (fn < 0) {
+      StringPushStringCharStar(buffer,"-");
+      fn = - fn;
+    }
     switch (unbox_fixnum(base)) {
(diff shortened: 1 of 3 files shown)


Bike:
I'm a little worried for how this works for `most-negative-fixnum`, since `(- most-negative-fixnum)` isn't a fixnum. (Or shouldn't be, I guess it somehow is, presumably due to a `-` problem.)

kpoeck:
Ser my own reported issue #579 :-)

kpoeck:
With the fix:
```lisp
(LET ((*PRINT-BASE* 16))
         (WITH-OUTPUT-TO-STRING (*STANDARD-OUTPUT*)
           (write most-negative-fixnum)))
-> "-2000000000000000"
````
which looks absolutely right, but let me see why it is printing this right

Bike:
Since the "fixnum" doesn't leave C++ it's probably okay. You're right.

PR #610, opened 2018-09-18
Fix #579 by special treating of most-negative-fixnum

### LINKED ISSUE #579: (- most-negative-fixnum) -> -2305843009213693952

kpoeck:
fixes #579 
before:
CL-USER> (- most-negative-fixnum)
-2305843009213693952
now 
COMMON-LISP-USER> (- most-negative-fixnum)
2305843009213693952

regression-test in numbers.lisp
(test negate-most-negative-fixnum-1
      (plusp (- -2305843009213693952)))


Bike:
I should probably replace the `negate` call unary `-` expands into with something inline,too

PR #611, opened 2018-09-18
In recursice calls to write object decrement print-level

kpoeck:
Now:
(let ((*print-level* 1))(write (list (list 1))))
-->
(#)   ;;;printed
((1)) ;;; returned
as in sbcl

was correct in ecl

### DIFF
--- a/src/core/write_list.cc
+++ b/src/core/write_list.cc
@@ -117,6 +117,8 @@ void Cons_O::__write__(T_sp stream) const {
     }
     y = oCar(x);
     x = oCdr(x);
+    // recursion with printlevel -1
+    DynamicScopeManager scope(cl::_sym_STARprint_levelSTAR, clasp_make_fixnum(print_level - 1));
     write_object(y, stream);
     /* FIXME! */
     if (!x || cl__atom(x) ||


Bike:
Only for conses, though? Maybe it should be independent of type? I mean, for other aggregate objects.

kpoeck:
IF other aggregate object recursively call write_object on their elements, yes.
grep -i -I -r "write_object" * gives like 1 page of calls, so let me check them

kpoeck:
Checked: For structs this is also wrong, e.g. 
```common-lisp
COMMON-LISP-USER> (defstruct test a)
TEST
COMMON-LISP-USER> (let ((*print-level* 2))
            (write-to-string  (make-test :a (list (list 1 2) 2))))
"#S(TEST :A ((1 2) 2))"
````
That should print "#S(TEST :A (# 2))"
Just fixed that, will send pr

Arrays do fine, see write_array_inner, where *print-level* is respected, e.g.
```lisp
(let ((*print-level* 1)) (write-to-string (make-array 2 :initial-contents (list 1 (list 1 2)))))
->
"#(1 #)"
````

PR #614, opened 2018-09-25
rename reader_error to reader_error_internal, don't expose to lisp

kpoeck:
To avoid ansi-error of having reader-error fbound


drmeister:
Thank you

PR #615, opened 2018-09-30
Respect *print-level* better in print-object for structure

kpoeck:
- if print-level = 0 behave for (destruct test) as for (defstruct test a)
- Decrement *print-level* in printing the slot-values
- complements #611

### DIFF
--- a/src/lisp/kernel/clos/print.lsp
+++ b/src/lisp/kernel/clos/print.lsp
@@ -186,11 +186,13 @@ printer and we should rather use MAKE-LOAD-FORM."
 (defmethod print-object ((obj structure-object) stream)
   (let* ((class (si:instance-class obj))
 	 (slotds (class-slots class)))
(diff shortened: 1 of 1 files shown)


Bike:
When you write these, please don't put in comments that only make sense if you're familiar with how the code previously went. It's not going to make any sense to a reader. If you do want to note the change, explain what the previous code was in the comment so the reader has context.

Other than that this seems fine. I think the original idea was this line in the clhs: 'If an object to be recursively printed has components and is at a level equal to or greater than the value of *print-level*, then the object is printed as "#".' A defstruct with no slots doesn't "have components", so this wouldn't apply.  But this is super anal and I think I'd rather just have structs print more consistently like this.

drmeister:
I've got to catch up reviewing your pull requests kpoeck - sorry about the delay.

drmeister:
Thank you for this.

Issue #558, opened 2018-07-28
ansi-tests: sxhash should return a positive fixnum


kpoeck:
```lisp
(typep (SXHASH (MAKE-SYMBOL "FOO")) 'fixnum)
NIL
(type-of (SXHASH (MAKE-SYMBOL "FOO")))
(INTEGER 8995143153023308406 8995143153023308406)
````
but in clhs http://www.lispworks.com/documentation/HyperSpec/Body/f_sxhash.htm
Function SXHASH
Syntax:
sxhash object => hash-code
Arguments and Values:
object---an object.
hash-code---a non-negative fixnum.

kpoeck:
cl__sxhash returns Integer_O::create(hg.hash());
hg.hash returns a  gc::Fixnum
a gc::Fixnum might not always fit in a lisp fixnum?

Bike:
a `gc::Fixnum` is, i think, an `int64_t`, so yeah you can fit non fixnums in there. Problematic. I guess we could like, shift out the low bits, or something...?

kpoeck:
This gives a runtime error in mcclim in incremental-redisplay, since there the result of sxhash is declared fixnum and obviously this is checked at runtime.

Will try to shift 2 bits to the right

kpoeck:
Should have it now correct will send pr

PR #628, opened 2018-11-04
Correct sxhash to spec, test for this problem and other hash-table-issues

kpoeck:
Solves #558 (retested with ansi-tests)

### DIFF
--- a/src/core/primitives.cc
+++ b/src/core/primitives.cc
@@ -1570,11 +1570,7 @@ CL_DEFUN Fixnum_sp cl__sxhash(T_sp obj) {
     return make_fixnum(1);
   HashGenerator hg;
   clasp_sxhash(obj, hg);
-  gc::Fixnum hash = hg.hash();
-  if (hash < 0)
-    hash = - hash;
-  if (hash > MOST_NEGATIVE_FIXNUM)
-    hash = hash >> 3;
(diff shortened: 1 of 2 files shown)


src/core/primitives.cc
```diff
@@ -1565,12 +1565,17 @@ CL_DEFUN T_sp cl__type_of(T_sp x) {
...
+CL_DEFUN Fixnum_sp cl__sxhash(T_sp obj) {
   if (obj.nilp())
     return make_fixnum(1);
   HashGenerator hg;
   clasp_sxhash(obj, hg);
-  return Integer_O::create(hg.hash());
+  gc::Fixnum hash = hg.hash();
+  if (hash < 0)
+    hash = - hash;
+  if (hash > MOST_NEGATIVE_FIXNUM)
```
drmeister: Won't this always be true given the if (hash <0) hash = -hash?

kpoeck: In current clasp:
(sxhash #*11) -> -355494570821960769
so I don't think that if (hash <0) hash = -hash holds in the current implementation

Bike: It might be better to figure out where the negative value is coming from.

drmeister:
Great!

PR #655, opened 2018-12-30
more details for recursive cl__error

kpoeck:
Works fine for me, pleas please review carefully whether dbg_safe_backtrace() can be used here


drmeister:
Yes - it should always be safe to execute dbg_safe_backtrace().
For example - I have dbg_safe_backtrace() being invoked when you hit control-T on macOS and when the system gets a SIGUSR2 on linux.

PR #656, opened 2018-12-30
make floor robust to nan arguments

kpoeck:
make floor not trap on 
* (floor (core:nan 23)) or if nan-fix  already merged on
* (floor (core:nan)) 

Not sure if this needed to be added to more math-functions

### DIFF
--- a/src/core/num_co.cc
+++ b/src/core/num_co.cc
@@ -186,24 +186,39 @@ Real_mv clasp_floor1(Real_sp x) {
   }
   case number_SingleFloat: {
     float d = x.unsafe_single_float();
-    float y = floorf(d);
-    v0 = _clasp_float_to_integer(y);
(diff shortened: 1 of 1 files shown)


drmeister:
Let's add support for these traps...
https://shinmera.github.io/float-features/#MACRO-FUNCTION%20FLOAT-FEATURES%3AWITH-FLOAT-TRAPS-MASKED

kpoeck:
Actually I think in clasp these traps are masked, that why we are getting the nans

drmeister:
But should we be letting NAN's propagate through the code?
Doesn't that mean we have to modify all of the numeric routines to handle NAN?
Is it better to signal an error the first time?

drmeister:
I'll merge this one now - but we will have to revisit this later.

PR #658, opened 2018-12-30
Fix cl__random error type

kpoeck:
Now:
```lisp
COMMON-LISP-USER> (random 0)
Condition of type: TYPE-ERROR
0 is not of type (OR (INTEGER 1) (FLOAT (0.0))).
````
before the float limit - my error - was 1


drmeister:
Thank you!

PR #661, opened 2019-01-06
Verify first argument of  special-operator-p to be a symbol

kpoeck:
In bclasp the is a call to special-operator-p with forms being lists, did fix this, otherwise the build fails


drmeister:
I merged this one manually - I'll not close this until I'm sure it's incorporated.

drmeister:
Done - thank you.

PR #671, opened 2019-01-12
Make compile-file-parallel more robust, signal error in the main thread

kpoeck:
Please review carefully if you agree to this
* solves error-handling compiling the regression suite

* Catches the error that might occur in the thread created for compile-from-ast.
* Signals the error in the main thread
* Only signals the first error
* Stack from error thread is not stored


drmeister:
Great! Thank you.

PR #688, opened 2019-02-10
Delete the temporay files left behind by parallel compiler

kpoeck:
test e.g. with 
````
(time (load "sys:regression-tests;run-all.lisp"))
````
and verify that no temp directories pile up


drmeister:
I thought I would deal with this by not generating object files for each form in the first place.
But that would mean generating a new kind of fasl file of concatenated object files - which is kind of a bad idea.
Let me think on this one a bit more.

kpoeck:
I don't see how I can upgrade this pr to current code, so I will resubmit a new one

PR #691, opened 2019-02-10
Fix format printing negative -0.0

kpoeck:
See tests
*  print-13
*  print-14


kpoeck:
what you changed is obviously fine. How do I proceed from here?

Bike:
I was holding off merging because I had a big merge I wanted to do, but I can't so let's just do this

PR #700, opened 2019-02-16
Fix constituent invalid trait & reader macros

kpoeck:
* Use the lisp version of sharp-a if defined
* Use the lisp version of sharp-s if defined 
* make the above 2 symbols known
* check for invalid traits in symbolTokenStr
* correct some types that gave errors in the lisp->c++ conversion
* return reader_errors and not simple-errors when possible
* regression tests for all

### DIFF
--- a/src/lisp/regression-tests/read01.lisp
+++ b/src/lisp/regression-tests/read01.lisp
@@ -52,6 +52,59 @@
 (test error-mcclim-1
       (list :\.))
 
+(test issue-382-a
+      (arrayp (let ((*readtable* (copy-readtable nil)))
+                (read-from-string "#a(T 1 (2))"))))
+
+(test issue-382-b
+      (arrayp (let ((*readtable* (copy-readtable nil)))
+                (read-from-string "#a(T (1) (2))"))))
+
+(test issue-382-c
+      (arrayp (let ((*readtable* (copy-readtable nil)))
+                (read-from-string "#a(T (2 2) ((1 2)(3 4)))"))))
+
+(test issue-382-d
+      (arrayp (let ((*readtable* (copy-readtable nil)))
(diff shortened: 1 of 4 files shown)


drmeister:
It looks great!  Thank you.

src/lisp/regression-tests/read01.lisp
```diff
@@ -52,6 +52,59 @@
...
+      (arrayp (let ((*readtable* (copy-readtable nil)))
+                (read-from-string "#2A((0 1 5) (foo 2 (hot dog)))"))))
+
+(defstruct %foo% %bar%)
+
+(test sharp-s-new-readtable
+      (let ((*readtable* (copy-readtable nil)))
+        (%foo%-p (read-from-string "#S(%FOO% :%BAR% 1)"))))
+
+(test 'SET-SYNTAX-FROM-CHAR-TRAIT-X-1a
```
phoe: This test name has a quote...

src/lisp/regression-tests/read01.lisp
```diff
@@ -52,6 +52,59 @@
...
+                             (LET ((*READTABLE* (COPY-READTABLE NIL)))
+                                (handler-case 
+                                   (READ-FROM-STRING
+                                    (CONCATENATE 'STRING (STRING C) "Z"))
+                                 (reader-error (e) e)))))
+                       chars))))
+        (and (typep (first result) 'reader-error)
+             (typep (eighth result) 'reader-error))))
+
+(test SET-SYNTAX-FROM-CHAR-TRAIT-X-1b
```
phoe: ...but this doesn't.

src/lisp/regression-tests/read01.lisp
kpoeck: The quote is not need, will delete it

PR #702, opened 2019-02-23
Add __write__ for ratios, did use method for symbols

kpoeck:
Before:
```lisp
(let ((*read-base* 10)
        (ratio (read-from-string "2140/969")))
            (let ((*print-base* 4))
               (write-to-string ratio)))
"2140/969"
````
after
```lisp
(let ((*read-base* 10)
        (ratio (read-from-string "2140/969")))
            (let ((*print-base* 4))
               (write-to-string ratio)))

"201130/33021"

````

### DIFF
--- a/include/clasp/core/numbers.h
+++ b/include/clasp/core/numbers.h
@@ -726,6 +726,9 @@ namespace core {
     bool minusp_() const {
       return clasp_minusp(this->_numerator);
     }
+
+    virtual void __write__(T_sp strm) const;
+    
     Ratio_O() : _numerator(clasp_make_fixnum(0)), _denominator(clasp_make_fixnum(1)) {};
     virtual ~Ratio_O() {};
   };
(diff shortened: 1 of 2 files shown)


kpoeck:
Now fixes also
```lisp
(string=
       "#10r-7/16"
       (let ((num -7/16)
             (*PRINT-READABLY* T)
             (*PRINT-ARRAY* NIL) (*PRINT-BASE* 10) (*PRINT-RADIX* T)
             (*PRINT-CASE* :CAPITALIZE) (*PRINT-CIRCLE* NIL)
             (*PRINT-ESCAPE* NIL) (*PRINT-GENSYM* NIL) (*PRINT-LEVEL* 47)
             (*PRINT-LENGTH* 25) (*PRINT-LINES* 48) (*PRINT-MISER-WIDTH* NIL)
             (*PRINT-PRETTY* NIL) (*PRINT-RIGHT-MARGIN* NIL)
             (*READ-DEFAULT-FLOAT-FORMAT* 'DOUBLE-FLOAT)
             )
         (write-to-string num)))
````

PR #704, opened 2019-02-24
defclass: make sure that defclass option "NIL" errors

### LINKED ISSUE #703: defclass accepts invalid options
Steps to reproduce:

(defclass xxx nil nil nil)

should error (nil is not valid initarg). Originally reported here: https://gitlab.com/embeddable-common-lisp/ecl/issues/474 .

dkochmanski:
This is as for CLHS section 7.1.2. Error was caused by the fact that
unknown-key was a flag, so if the initarg was NIL, we were assigning
it value NIL, what is also a boolean false. Right now we collect all
invalid initargs in a list, so in case of NIL we'll get (NIL) what is
a generalized boolean true value. Closes #703.

Note that I didn't compile clasp to test this change. It takes too long
for me and I don't have a complete set of dependencies.

### DIFF
--- a/src/lisp/kernel/clos/standard.lsp
+++ b/src/lisp/kernel/clos/standard.lsp
@@ -636,11 +636,11 @@ because it contains a reference to the undefined class~%  ~A"
     (do* ((name-loc initargs (cddr name-loc))
 	  (allow-other-keys nil)
 	  (allow-other-keys-found nil)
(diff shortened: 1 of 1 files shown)


kpoeck:
Am testing your change

kpoeck:
Fixes the problem and shows no errors either in regression-tests nor in ansi-tests

PR #705, opened 2019-03-02
Finalize lisp interface of weakHashTable + regression-tests

kpoeck:
Did fix the lisp-interface of weak-hash-table (were all failing)
See src/lisp/regression-tests/hash-tables0.lisp for a bit more complete set of tests
Current code now has no failures in the test-suite of cl+ssl (need my pull request for cffi though)

Attention:
* /src/lisp/kernel/clos/boot.lsp depends on src/lisp/kernel/clos/hierarchy.lsp because of a constant
* that does not seem to be taken in by the build process

### DIFF
--- a/src/core/hashTable.cc
+++ b/src/core/hashTable.cc
@@ -229,7 +229,7 @@ HashTable_sp HashTable_O::create_thread_safe(T_sp test, SimpleBaseString_sp read
   return ht;
 }
 
-T_mv HashTable_O::maphash(T_sp function_desig) {
+void HashTable_O::maphash(T_sp function_desig) {
     //        printf("%s:%d starting maphash on hash-table@%p\n", __FILE__, __LINE__, hash_table.raw_());
   Function_sp func = coerce::functionDesignator(function_desig);
   HT_READ_LOCK(this);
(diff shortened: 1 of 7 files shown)


src/core/hashTable.cc
```diff
@@ -246,21 +246,20 @@ T_mv HashTable_O::maphash(T_sp function_desig) {
 CL_LAMBDA(function-desig hash-table);
 CL_DECLARE();
 CL_DOCSTRING("see CLHS");
-CL_DEFUN T_mv cl__maphash(T_sp function_desig, T_sp thash_table) {
+CL_DEFUN T_mv cl__maphash(T_sp function_desig, HashTableBase_sp hash_table) {
```
drmeister: Could you change this to return a T_sp?   T_mv is for Values(...) but T_sp is more correct as you point out.

kpoeck: I only correct half of the problem as you spotted will do as you said and look for other occurances

kpoeck:
I did implement your suggestion, please check again

drmeister:
Thank you very much!

PR #710, opened 2019-03-16
CHANGED: Make PERCENTinc_pointer return new ForeignData_O object

dg1sbg:
As discussed on #clasp on 2019-03-16.


drmeister:
Thank you and that was awesome.
People helping people! :-)

PR #714, opened 2019-03-24
Fix equalp for string vs vector

### LINKED ISSUE #713: babel-test: equalp does not return
```lisp
(equalp 
  (make-array 4 :element-type 'character :initial-contents (list #\a #\SUB #\b #\c))
   #(#\a #\Sub #\b #\c))
````
does not return.
this is the simplified version of test 
```lisp
(defstest dec.ascii.3
    (octets-to-string (ub8v 97 255 98 99) :encoding :ascii :errorp nil)
  #(#\a #\Sub #\b #\c))
````

kpoeck:
Fixes #713 for babel


drmeister:
That is an obvious one - thank you.

PR #717, opened 2019-03-29
Let core__reader_double_quote_string return SimpleCharacterString_sp

kpoeck:
Solves the following in the repl:
```lisp
(string-equal
       (make-array 4 :element-type 'character :initial-contents (list #\? (code-char 256) #\? #\? ))
       (let ((var (copy-seq "????")))
         (setf (char var 1) (code-char 256))
         var))
T
````
With the pr var would be a simple-base-string and (setf (char var 1) (code-char 256)) would fail.

Unfortunately this only solves half the problem, since in cmpliteral.lsp, strings are treated with ltv/base-string, which will again convert strings to simple-base-string.

I seem to lack know-how on the c++ side, so I do this as a first part to solve the problem.

Here advise from drmeister on #clasp 
::notify kpoeck I'd use ltvc_make_simple_string(gctools::GCRootsInModule* holder, size_t index, size_t string_length, const int* str) - you want to use unicode code points - right?  So an array of 4-byte integers for the code points would be way more than enough.  I'd have to show you how to add support for it to the new byte code intepreter - but it's easy enough.

### DIFF
--- a/include/clasp/core/array.h
+++ b/include/clasp/core/array.h
@@ -1389,6 +1389,7 @@ namespace core {
     bool all_base_char_p() const;
     /*! Return the smallest character simple-string that can hold this */
     SimpleString_sp asMinimalSimpleString() const final;
+    SimpleCharacterString_sp asSimpleString () const;
   };
 };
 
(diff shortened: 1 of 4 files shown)


kpoeck:
has problems with the code base where sometimes it is assumed that a string is a simple-base-string, so I close that for now

PR #718, opened 2019-03-29
Fix package-nicknames for :ast-tooling + test

### LINKED ISSUE #699: AST-TOOLING package has bad nickname
`(package-nicknames (find-package "AST-TOOLING")) => (NIL)`

It should of course be `()`. This is apparent all the way back to iclasp, so I assume it's part of the C++ package definition. However things look okay in my generated `symbols_scraped_inc.h` so I don't know what's up.

Package nicknames must be strings. This problem results in SLIME's `slime-repl-set-package` aka `,i` command not working with Clasp.

kpoeck:
Fixes #699 
see tests:
* issue-699
* issue-699-generic

### DIFF
--- a/src/asttooling/clangTooling.cc
+++ b/src/asttooling/clangTooling.cc
@@ -731,7 +731,7 @@ void initialize_clangTooling() {
   initialize_matchers();
   // overloaded functions that had trouble resolving
   clang::ASTContext &(clang::ASTUnit::*clang_ASTUnit_getASTContext)() = &clang::ASTUnit::getASTContext;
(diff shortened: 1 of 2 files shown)


Bike:
What does this do, exactly? Remove CLANG as a nickname? Do we need that anywhere?

kpoeck:
I believe the definition of package is in https://github.com/clasp-developers/clasp/blob/dev/include/clasp/clbind/scope.h#L114, so yes, it removes the nickname. 

can't find any use of "CLANG" via grep -i -I -r "\"CLANG\"" *

Bike:
Guess it's fine then.

PR #734, opened 2019-04-21
Feature signum bignum

kpoeck:
Before
```lisp
COMMON-LISP-USER> (signum (1+ most-positive-fixnum))
../../src/gctools/hardErrors.cc:30:throw_hard_error_subclass_must_implement className: BIGNUM method: signum_
Abort trap: 6
````
now
```lisp
COMMON-LISP-USER> (signum (1+ most-positive-fixnum))
1
````

PR #735, opened 2019-04-22
Teach cst-read semi-standard-chars

kpoeck:
With that the regression suite compiles, e.g.
* strings01.lisp with  #\Nul
* character0.lisp

Please note that the regression suite also needs https://github.com/clasp-developers/clasp/pull/734 to execute (or define signum again in lisp

### DIFF
--- a/src/lisp/kernel/cleavir/translate.lisp
+++ b/src/lisp/kernel/cleavir/translate.lisp
@@ -696,17 +696,16 @@ This works like compile-lambda-function in bclasp."
                                  ("BELL"   . #.(code-char 7))
                                  ("ESCAPE" . #.(code-char 27))
                                  ("DEL"    . #.(code-char 127))
-                                 ("CR"     . #.(code-char 13))
-                                 ("RUBOUT" . #.(code-char 127)))
+                                 ("CR"     . #.(code-char 13)))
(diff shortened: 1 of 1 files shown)


Bike:
Is this not a matter for Eclector?

kpoeck:
Yes a matter for eclector, that's why I added eclector.reader:find-character for the client, which is - as far as I understand - the supported eclector interface for this, see https://github.com/robert-strandh/Eclector/blob/master/code/reader/generic-functions.lisp#L25

Bike:
What I meant was, would this not be better as a change to Eclector itself? But I see it's doing only the character names specifically defined in the standard, so never mind.

Then I think this is fine but with some little changes-

 * RUBOUT is already in the standard characters so we don't need to have it.
 * The Eclector docs specify that `find-character` gets an uppercase string, if I'm reading them correctly, so the `stringp` check and check for `#\u` shouldn't be necessary.
 * The result of `parse-integer` should be checked to not be null, e.g. if someone writes `#\uvula` - because that would end up doing `(<= #x00 nil #x10ffff)` which would be a problem.

kpoeck:
Understood, testing

kpoeck:
Changed as per your comments

PR #744, opened 2019-05-01
add -cst or -non-cst cl__lisp_implementation_version()

kpoeck:
(lisp-implementation-version) ->
* "cclasp-boehm-0.4.2-813-ga2b37b3fe-non-cst" or
* "cclasp-boehm-0.4.2-822-gcc308e33b-cst"

Helps to distinguish the version in cl-bench and the fasls in quicklisp .cache (i have cst and non-cst builds at the same time)


drmeister:
Do we need the non-cst anymore?   I would advocate that we throw the switch and make cst the only version of clasp.

kpoeck:
I cannot decide on whether we need non-cst, just wanted to have something that distinguishes both build for cl-bench

PR #749, opened 2019-05-09
Fix progv as per ansi-tests

kpoeck:
Before
```lisp
COMMON-LISP-USER> (let ((a 1))(declare (special a))(progv '(a) nil a))
1
````
After
```lisp
COMMON-LISP-USER> (let ((a 1))(declare (special a))(progv '(a) nil a))
Condition of type: UNBOUND-VARIABLE
The variable A is unbound.
Available restarts:
(use :r1 to invoke restart 1)
````

In the scope of the progv a must be unbound, because no value was passed

### DIFF
--- a/src/core/compiler.cc
+++ b/src/core/compiler.cc
@@ -1029,10 +1029,14 @@ CL_DEFUN T_mv core__progv_function(List_sp symbols, List_sp values, Function_sp 
   int index = 0;
   for (auto curSym : symbols) {
     if (index < lengthValues) {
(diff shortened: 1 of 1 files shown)


Bike:
I'm a little antsy about this. The relevant CLHS line is `If too few values are supplied, the remaining symbols are bound and then made to have no value.` which is kind of gibberish.

drmeister:
Ill let Bike weigh in on this one.

Bike:
I guess I don't have any alternate idea of how it should work.

PR #762, opened 2019-05-27
Correct Packages in ext:make-encoding

kpoeck:
load-encoding is in package ext


kpoeck:
Drmeister did 50% of this change independently, let me find out how to resolve that

PR #767, opened 2019-05-28
Trivial: Feature more character names for cst reader

kpoeck:


Bike:
Eclector already has RUBOUT. Is there a reason for these characters specifically, otherwise?

kpoeck:
#\Sub is used in Babel tests, for #\Rubout I than made an error, will delete that

PR #778, opened 2019-05-29
Attempt at fixing FD_ZERO() and FD_COPY() issues / differences …

dg1sbg:
…between macOS and Linux (Debian).


drmeister:
Oh - that's nicer than mine - thank you!

Issue #764, opened 2019-05-28
Regression: Structure access hangs when non struct passed as argument


kpoeck:
Consider
```lisp
Starting Boehm cclasp 0.4.2-968-g7c9f134bc ... loading image...
Top level in: #<PROCESS TOP-LEVEL @0x1111dfc39>.
COMMON-LISP-USER> (declaim (optimize (speed 0)(safety 3)))
((CLASP-CLEAVIR::MAINTAIN-SHADOW-STACK . T) (CLASP-CLEAVIR::DO-TYPE-INFERENCE)
 (CLASP-CLEAVIR::DO-DX-ANALYSIS) (CLEAVIR-ESCAPE:TRUST-DYNAMIC-EXTENT)
 (CLEAVIR-KILDALL-TYPE-INFERENCE:INSERT-TYPE-CHECKS . T))
COMMON-LISP-USER> (defstruct foo bar)
FOO
COMMON-LISP-USER> (foo-bar 0)

````
and clasp hangs.

kpoeck:
lldb shows:
````
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x17)
    frame #0: 0x000000011213716d
->  0x11213716d: movq   0x17(%rbx), %rax
    0x112137171: andq   $-0x8, %rax
    0x112137175: movq   0x18(%rax), %rax
    0x112137179: movq   0x110(%rip), %rcx
Target 0: (clasp) stopped.

Executable module set to "/Users/karstenpoeck/lisp/compiler/clasp-karsten/build/clasp".
Architecture set to: x86_64-apple-macosx-.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x17)
  * frame #0: 0x000000011213716d
    frame #1: 0x000000010739ac99 clasp`core::cl__apply(head=<unavailable>, args=<unavailable>) at evaluator.cc:0:49 [opt]
    frame #2: 0x00000001041846f8 clasp`APPLY^COMMON-LISP^FN^^ at cl-wrappers.lisp:1915:1
    frame #3: 0x0000000106455393 clasp`SIMPLE-EVAL^CLASP-CLEAVIR^FN^^ at toplevel.lisp:51:15
    frame #4: 0x000000010645a24c clasp`CCLASP-EVAL^CLASP-CLEAVIR^FN^^ at toplevel.lisp:126:3
[...]

PR #786, opened 2019-06-03
Applied stassats proposals

kpoeck:
Basically from stassats on irc clasp (I deleted the names)
```lisp 18:25:50
can you edit handle_or_queue?
to go directly to handle_signal_now when signal_code is sigsegv 
sigbus shouldn't be queued either 
add sigill as well
````

Test as follows
```lisp
(defstruct foo bar)

FOO
COMMON-LISP-USER> (foo-bar 0)
../../src/gctools/interrupt.cc:160 Received a unix signal with code: 11

Condition of type: UNIX-SIGNAL-RECEIVED
Serious signal 11 caught.

Available restarts:
(use :r1 to invoke restart 1)

1. (CONTINUE) Ignore signal
````
Before this change that used to loop, see e.g. #764

### DIFF
--- a/src/gctools/interrupt.cc
+++ b/src/gctools/interrupt.cc
@@ -233,10 +233,12 @@ void handle_all_queued_interrupts()
 }
 
 void handle_or_queue(core::ThreadLocalState* thread, core::T_sp signal_code ) {
-  if (!signal_code) {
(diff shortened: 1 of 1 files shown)


drmeister:
Sounds good!

PR #788, opened 2019-06-07
on verbose show time and bytes spent compiling the file

kpoeck:
-> e.g.
--60936(out)--> Time run(241.352 secs) consed(1071982232 bytes)
so that slow compiles can be seen


drmeister:
Looks good - thank you.

PR #802, opened 2019-06-22
Fix filter vector for displaced arrays

### LINKED ISSUE #801: Regression ansi-tests: remove* not working correctly on displaced arrays
```lisp
(let ((s (make-array 10 :element-type 'base-char
                     :displaced-to (make-array 15 :initial-contents "XXab1c23def4YYY" :element-type 'base-char)
                     :displaced-index-offset 2)))
  (remove-if #'ALPHA-CHAR-P s :count 3))
->
"1def4"
but should be "123def4"
````

kpoeck:
Fixes #801

### DIFF
--- a/src/lisp/kernel/lsp/seqlib.lsp
+++ b/src/lisp/kernel/lsp/seqlib.lsp
@@ -194,7 +194,7 @@
                         start (1+ start)))
               (incf untransformed-index)))
           ;; ... and copy the elements outside the limits
-           (copy-subarray out start in end l)
+          (copy-subarray out start in end l)
           (values out (+ start (- l end))))))))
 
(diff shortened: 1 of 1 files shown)


Bike:
`index` is newly bound by do-subvector, so you just need to put in a new variable for that and use the original index as the "untransformed" index.

That said, this is tight code and I don't want a second incf that's obviously redundant... we should work out something better.

Bike:
We can rewrite do-subvector to bind the index as a symbol macro that is just underlying-index + displaced-index-offset. I think that should cover all uses of it?

kpoeck:
Are you doing this or should I try to implement what you proposed?

Bike:
If you could do it that would be cool.

kpoeck:
ok

PR #807, opened 2019-06-24
Fix wrong ftype in define-list-struct-accessors

### LINKED ISSUE #804: defstruct generates incorrect ftype in accessors for type :list
to reproduce:
-- need ast build (since cst does not seem to do this type check
```lisp
(declaim (optimize (safety 3)))
(DEFSTRUCT (STRUCT-TEST-FOO1 (:TYPE LIST)) bar)

(LET ((TEMP-5 (MAKE-STRUCT-TEST-FOO1 :bar 23)))
  (STRUCT-TEST-FOO1-BAR TEMP-5))
````
-> 
STRUCT-TEST-FOO1 is not a valid type specifier.
   [Condition of type SIMPLE-ERROR]
````

kpoeck:
Fixes #804

### DIFF
--- a/src/lisp/kernel/lsp/defstruct.lsp
+++ b/src/lisp/kernel/lsp/defstruct.lsp
@@ -359,7 +359,7 @@
                                     nil
                                     `((define-setf-expander ,accname (object &environment env)
                                         (get-setf-expansion (list 'nth ,index object) env))))))
-                         (list* `(declaim (ftype (function (,name) ,type) ,accname) ; useless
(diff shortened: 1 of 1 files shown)


Bike:
Uhh, why is it dummied out

if this is a list struct it should be type `list`. also, despite my earlier comment, it's not useless in that it should at least trigger a warning if called with the wrong number of parameters

kpoeck:
ok, will send update implementing your proposal (and dropping the comment)

kpoeck:
Hope that I introduced your comment, passes all tests in ansi-tests

PR #809, opened 2019-06-29
Special case literal::immediate-datum

### LINKED ISSUE #797: Ansi-tests: random tester:#(LITERAL::IMMEDIATE-DATUM -514012002144732176) is not of type INTEGER.
```lisp
COMMON-LISP-USER> (compile nil                                                           
         '(lambda()
           (LOAD-TIME-VALUE
            (FLET ((%F4 (&OPTIONAL &KEY (KEY1 9911946289546) (KEY2 -19296971321001))
                     3904101166444))
              -128503000536183044))))
; caught ERROR:
;   #(LITERAL::IMMEDIATE-DATUM -514012002144732176) is not of type INTEGER.

Condition of type: TYPE-ERROR
#(LITERAL::IMMEDIATE-DATUM -514012002144732176) is not of type INTEGER.

Available restarts:
(use :r1 to invoke restart 1)

Broken at frame[0] NIL. In: #<PROCESS TOP-LEVEL @0x11bce64b9>.
; caught WARNING:
;   compiled-function-file expected a function as argument - but it got NIL - there may not be any backtrace available

````

kpoeck:
Fixes #797

### DIFF
--- a/src/lisp/kernel/cleavir/translate.lisp
+++ b/src/lisp/kernel/cleavir/translate.lisp
@@ -112,10 +112,15 @@ when this is t a lot of graphs will be generated.")
   (let ((defs (cleavir-ir:defining-instructions location)))
     (= (length defs) 1)))
 
+(defun translate-immediate-input-value (value)
+  (if (literal::immediate-datum-p value)
+      (literal::immediate-datum-value value)
+      value))
+
 (defun in (datum &optional (label ""))
   (cond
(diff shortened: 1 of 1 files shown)


kpoeck:
Did run the ansi-tests to verify

Bike:
I think we should just not have a literal datum there- haven't looked at it in detail yet, though

kpoeck:
fix by drmeister in another way

PR #812, opened 2019-07-06
consider clasp_print_readably in BitVector_O::__write__ and SimpleBitVector_O::__write__

kpoeck:
so that
```lisp
(let ((*print-array* nil)
      (*print-pretty* nil)
      (*PRINT-READABLY* t))
  (with-output-to-string (stream)
      (princ (make-array 2 :element-type 'bit :initial-element 0) stream)))
````
results in
```lisp
"#*00"
````
and not in
```lisp
"#*"
````

### DIFF
--- a/src/core/write_array.cc
+++ b/src/core/write_array.cc
@@ -197,41 +197,37 @@ namespace core {
 #endif
 
     void SimpleBitVector_O::__write__(T_sp stream) const {
-	if (!clasp_print_array() && !clasp_print_readably()) {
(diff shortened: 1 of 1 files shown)


Bike:
makes sense to me, except isn't the second "if" redundant? the `(or *print-array* *print-readably*)` one. and the indentation seems off

but that's nitpicking, fundamentally this is ready for merge

kpoeck:
The indentations is what emacs gives, so I don't know what to fix there.
I'll deletet the second if

PR #816, opened 2019-07-09
Test for getcwd-null-pointer

kpoeck:
need when another bug did not close the file descriptors and :getcwd(NULL,0); failed


Bike:
looks basically fine to me, but isn't the `ok && nowpwd` check redundant? i mean, if `!ok`, it signals an error, so `ok` will never be NULL in the second test

kpoeck:
Yes, is redundant, fixing

PR #818, opened 2019-07-12
Fix ext  getcwd null pointer

kpoeck:


kpoeck:
for whatever reason I coudn't update #816

Issue #595, opened 2018-09-02
cl-bench: conversion to floats of ratios with numerator/denominator > largest float -> #<single-float quiet NaN>


kpoeck:
```lisp
COMMON-LISP-USER> (float (/ (expt 10 300)(1+ (expt 10 300))))
1.0
COMMON-LISP-USER> (float (/ (expt 10 400) (1+  (expt 10 400))))
#<single-float quiet NaN>
````
in ecl
```lisp
> (float (/ (expt 10 300)(1+ (expt 10 300))))
1.0
> (float (/ (expt 10 400) (1+  (expt 10 400))))
1.0
> most-positive-double-float
1.7976931348623157d308
````

kpoeck:
Seems to ecl has a more intelligent routine to convert a ratio to a float
```c
static double
ratio_to_double(cl_object num, cl_object den)
{
  cl_fixnum scale;
  cl_object bits = prepare_ratio_to_float(num, den, DBL_MANT_DIG, &scale);
#if (FIXNUM_BITS-ECL_TAG_BITS) >= DBL_MANT_DIG
  /* The output of prepare_ratio_to_float will always fit an integer */
  double output = ecl_fixnum(bits);
#else
  double output = ECL_FIXNUMP(bits)? ecl_fixnum(bits) : _ecl_big_to_double(bits);
#endif
  return ldexp(output, scale);
}
````

kpoeck:
solved in #997

PR #827, opened 2019-07-20
[Wip] Fix log for bignums not to overflow

kpoeck:
Allows the following to work
```lisp
(labels ((factorial (n)(if (<= n 1) 1 (* n (factorial (1- n))))))
        (log (factorial 200)))

-> 863.232

````
w/o this pr #.ext:single-float-positive-infinity is returned

Also allows 
* pi-decimal/big
* pi-atan

to run. failed before


kpoeck:
Fixes #595

Bike:
Why is there a loop? Could the algorithm be more explained?

kpoeck:
The algorithm was taken from
https://gitlab.com/embeddable-common-lisp/ecl/blob/5f71f728a32d87b561aac1b5e21e8418e7edf447/src/c/number.d#L653

and adapted to the clasp coding convention. I see that in the meantime ecl has change its algorithm so let me check that

PR #828, opened 2019-07-20
Speed-up log* functions for fixnums

kpoeck:
Speed up cl-bench test crc40 from cl-bench

```lisp
(declaim (inline crc-division-step))

(defun crc-division-step (bit rmdr poly msb-mask)
  (declare (type (signed-byte 56) rmdr poly msb-mask)
	   (type bit bit))
  ;; Shift in the bit into the LSB of the register (rmdr)
  (let ((new-rmdr (logior bit (* rmdr 2))))
    ;; Divide by the polynomial, and return the new remainder
    (if (zerop (logand msb-mask new-rmdr))
	new-rmdr
	(logxor new-rmdr poly))))

(defun compute-adjustment (poly n)
  (declare (type (signed-byte 56) poly)
	   (fixnum n))
  ;; Precompute X^(n-1) mod poly
  (let* ((poly-len-mask (ash 1 (1- (integer-length poly))))
	 (rmdr (crc-division-step 1 0 poly poly-len-mask)))
    (dotimes (k (- n 1))
      (setf rmdr (crc-division-step 0 rmdr poly poly-len-mask)))
    rmdr))

(defun calculate-crc40 (iterations)
  (declare (fixnum iterations))
  (let ((crc-poly 1099587256329)
	(len 3014633)
	(answer 0))
    (dotimes (k iterations)
      (declare (fixnum k))
      (setf answer (compute-adjustment crc-poly len)))
    answer))

(defun run-crc40 ()
  (calculate-crc40 10))
````

speedup from 108.13 to 28.86 secs
(* rmdr 2) is now slowest part (+ rmdr rmdr) would be much faster

### DIFF
--- a/src/core/numbers.cc
+++ b/src/core/numbers.cc
@@ -246,20 +246,237 @@ CL_DEFUN Real_sp cl__max(Real_sp max, List_sp nums) {
   return max;
 }
 
+// logand, logxor, logior, logeqv are all nearly the same, apart from the operation
+// logandc1_id until lognor_id as well
+
+typedef enum {
+    logand_id = 0,
+    logxor_id = 1,
+    logior_id = 2,
+    logeqv_id = 3,
+    logandc1_id = 4,
+    logandc2_id =5,
+    logorc1_id = 6,
+    logorc2_id = 7,
+    lognand_id = 8,
+    lognor_id = 9
+} log_operations;
+
+Integer_sp log_operation_2op(log_operations operation, Integer_sp first, Integer_sp second) {
+  // if the arguments are all fixnum, don't convert everything to mpz, but stay in fixnums
+  if (first.fixnump() && second.fixnump()){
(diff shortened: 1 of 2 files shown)


Bike:
Could you summarize these changes? What optimization you did here?

kpoeck:
The old code converted every argument toba bignum and than applied the bignum operatios.

My change applies the fixnum log* operations if all arguments are fixnums. The fixnum formulaes are taken from ecl, as is the optimization idea

Bike:
OK, i see. that seems fine. however, this op enum thing seems odd to me. It's like `boole`. And we already have a boole ops enum: https://github.com/clasp-developers/clasp/blob/dev/src/core/bits.cc#L114-L150

Might it make more sense to have individual functions? logxor2 or whatever?

Bike:
I'm going to incorporate this but make changes later.

PR #830, opened 2019-07-21
Fix core__unix_get_local_time_zone and implicitly decode-universal-time

### LINKED ISSUE #829: (core:unix-get-local-time-zone) return unnormalized ratio
```lisp
CL-USER> (core:unix-get-local-time-zone)
-2/1
````
Should probably use Rational_O::create

kpoeck:
Fixes #829

### DIFF
--- a/src/core/unixfsys.cc
+++ b/src/core/unixfsys.cc
@@ -1734,7 +1734,7 @@ CL_DEFUN bool core__unix_daylight_saving_time(Integer_sp unix_time) {
 CL_LAMBDA();
 CL_DECLARE();
 CL_DOCSTRING("unixGetLocalTimeZone");
-CL_DEFUN Ratio_sp core__unix_get_local_time_zone() {
+CL_DEFUN Rational_sp core__unix_get_local_time_zone() {
   gctools::Fixnum mw;
 #if 0 && defined(HAVE_TZSET)
   tzset();
@@ -1752,8 +1752,11 @@ CL_DEFUN Ratio_sp core__unix_get_local_time_zone() {
     mw -= 24 * 60;
   else if (gtm.tm_wday == (ltm.tm_wday + 1) % 7)
     mw += 24 * 60;
+  // Fix from ecl
+  if (ltm.tm_isdst)
+    mw += 60;
 #endif
-  return Ratio_O::create(make_fixnum(mw), make_fixnum(60));
+  return Rational_O::create(make_fixnum(mw), make_fixnum(60));
 }
 
 CL_LAMBDA(dir mode);


Bike:
Could you change the "fix from ecl" comment to say what the actual problem is? The DST thing. "fix from ecl" isn't very helpful.

kpoeck:
Literally the instruction after the comment was meant:
// Fix from ecl
  if (ltm.tm_isdst)
    mw += 60;

the if (ltm.tm_isdst) mw += 60; 
was missing, so the time was 1 hour in the future.

after the fix (core:unix-get-local-time-zone) returns -1 and not -2/1

Bike:
If we were missing a plus sixty minutes, wouldn't our time be an hour in the past?

If you mean that without this change we're not accounting for DST at all, in the future just say that in a commit message, you know? But that doesn't match what you said in the issue - that seems to be saying we're accounting for DST twice? I don't understand and I want the code to be clear. The comment "fix from ECL" isn't going to help us in the future. It's not helping me right now, even.

kpoeck:
Let me start by quoting the definition of time zone from clhs:
time zone:
n.  a rational multiple of 1/3600 between -24 (inclusive) and 24 (inclusive) that represents a time zone as a number of hours offset from Greenwich Mean Time. Time zone values increase with motion to the west, so Massachusetts, U.S.A. is in time zone 5, California, U.S.A. is time zone 8, and Moscow, Russia is time zone -3. (When "daylight savings time" is separately represented as an argument or return value, the time zone that accompanies it does not depend on whether daylight savings time is in effect.)

So slightly surprisingly what in utc is said as utc+3, is counted in this definition as -3. I live in Frankfurt which is UTC+1, so offset -1 from GMT. In your case it will be a positive number (probably 5)

GMT by definition has no dst, but the local time can have dst (as it is in my case), right now - just looking at the hours gtm is 20 and ltm is 22. But not counting the dst, the difference is just 1 hour, so somehow we have arrive at -1 as timezone instead of -2.

now in the formula mw is defined as gtm in minutes - ltm in minutes (giving -120). Adding 60 min to compensate for dst in tlme gives -60. Which than divided by 60, gives -1, the correct answer.

Ecl fixed this dst handling for ticket https://gitlab.com/embeddable-common-lisp/ecl/issues/495

Bike:
OK, so is the issue here the `When "daylight savings time" is separately represented as an argument or return value, the time zone that accompanies it does not depend on whether daylight savings time is in effect.` bit? Right now we're returning the time zone regardless of DST, but we need to add the DST compensation to the returned time zone even though there's also a DST flag, and that's what this change does?

I can understand that much, hopefully. Now in the particular ECL change, the commit name is "si::get-local-time-zone: don't apply shift for daylight saving time", but then the actual code change seems to include an applied shift? I guess that can just be ECL being weird. It also says `the functions calling si::get-local-time-zone later do the necessary fixup for Daylight Saving Time on their own`, but in Clasp only `decode-` and `encode-universal-time` seem to and I don't see them doing any fixup. Is it perhaps true that this commit message is incorrect, and that what the change is doing is applying DST when it was not previously applied at all?

Sorry for going on about this, but I don't want to replace one mistake with another

kpoeck:
I can only speculate on the reasons for ecls message, but let me confirm the idea.
GMT never has dst by definition, but ltm (localtime) might have, and right now in Europe it really has.

If the distance between gmt and ltm is calculated w/o respecting dst, the difference might be 1 hour (dst goes forward) to big, resulting in a timezone that is smaller (more negative) as it should be. In case of Frankfurt, this results in -2 instead of -1, when dst is active (like now).

The real unfortunate side effect of this error, is that files created by clasp are right now created with a file-creation-date in the future, e.g. a file created at 19.00 has a timestamp of 20.00 (omitting all other time-fields). This is a catastrophe for all build-utilities comparing file-write-date of a source file (probably right date, since emacs doesn't have this error) and fasl file being 1 hour in the future. 

decode-universal-time does an adjustment for dst, see  
https://github.com/clasp-developers/clasp/blob/master/src/lisp/kernel/lsp/mislib.lsp#L208.

If (daylight-saving-time-p ...) returns true, the loop is entered a second time with tz-p and dstp set to true. 

This results in 
(decf ut (round (* (+ tz (if dstp -1 0)) 3600))), meaning ut getting decreased 1 hour.

TLDR; core:unix-get-local-time-zone ignores dst, but shoudn't

Bike:
I don't get it. Whatever. I'll just merge this. It doesn't break anything. Time sucks.

PR #834, opened 2019-07-27
Cleanup + Fix Regression Test read-from-string0

kpoeck:


Bike:
Please, like, tell us what these changes are?

In particular I don't get the lispReader one. The CLHS page you cite, 2.1.4.2, says that every character has "one or more" constituent traits, and it looks like with your change some characters won't have any.

kpoeck:
In the table in 2.1.4.2 it says the following 8 chars are of trait invalid:
Backspace, Tab, Newline, Linefeed     , Page         , Return , Space        , Rubout       

The question remains what happens to the characters that are not in the list. Obviously I convinced myself that not setting trait invalid for the other characters is the right thing to do. This decision had the nice consequence that the test read-from-string0 worked.

In this specific chapter the CLHS seems to be a bit broken so let me use the Franz.doc.

2.1.4.2 says:
"Every character has one or more constituent traits that define how the character is to be interpreted by the Lisp reader when the character is a constituent character. "

I understood this that only for constituent character constituent traits  is defined.

In https://franz.com/support/documentation/10.1/ansicl/subsecti/characte.htm
, figure 2.1.4 for the characters in the standard syntax it is marked if they are constituent or not.

Nul or NULL is clearly not in the list so I'd argue that Constituent Traits of NULL is undefined but not invalid
.
The best approximation for this I came up with was the change that we discuss:
```c++
  if (x == BACKSPACE_CHAR || x == TAB_CHAR || x == NEWLINE_CHAR || x == LINE_FEED_CHAR || x == PAGE_CHAR || x == RETURN_CHAR || x == ' ' || x == RUBOUT_CHAR)
    return (TRAIT_INVALID | x);
  else return (x);
````

Perhaps the next question is whether character that are not constituent can be part of a token. or better whether this test is valid:
```lisp
(test read-from-string0
      (eq
       (read-from-string (concatenate 'string "abc" (string #\nul) "AA"))
       (intern (concatenate 'string "ABC" (string #\nul) "AA"))))
````

Do you consider this test valid?
Do you follow my line of thought?

kpoeck:
for whats its worth:
```lisp
* (ECLECTOR.READER:READ-FROM-STRING (concatenate 'string "abc" (string #\nul) "AA"))
|ABCAA|
6
````

kpoeck:
and
```lisp
* (ECLECTOR.READER::CHAR-INVALID-P #\nul)
NIL
* (ECLECTOR.READER::CHAR-INVALID-P #\rubout)
T
````

Bike:
OK, so first off, if what you want to fix is clasp's result in the test, please just say that in the PR - what the problem is, what you did to fix it.

As for the test, I don't think we need it. I would expect NUL to be invalid (though I don't feel strongly about this) so the read ought to signal an error. The test is probably in there from the bad old days when clasp would freak out if a NUL was in a string anywhere, for C string reasons.

What I'm worried about with your change is a character having no constituent traits. I feel like this would cause weird issues if a token is supposed to be constructed out of it, i.e., in `interpret_token_or_throw_reader_error` and `symbolTokenStr` and stuff. It seems like by default we treat it as a character suitable for a symbol, but I don't know. How we deal with this now is probably not right - e.g. I would expect the various Latin script copies in higher Unicode to be alphabetic, but it looks like right now we treat them as invalid - but I'd prefer specific errors.

kpoeck:
I feel a but uneasy with your reasoning, especially since it does not follow what eclector does. 

Is there any place in the spec that justifies treating characters not in the table in https://franz.com/support/documentation/10.1/ansicl/subsubse/constit0.htm?

Would it be a compromise to treat characters not in this list with a new trait TRAIT_UNDEFINED? Additionally we need to specify, if these characters can be part of a token.

Bike:
You've submitted patches to Eclector yourself, haven't you? It's not a perfect reference implementation.

The spec - that page - says "Every character has one or more constituent traits". Every character. Not "every character in this table".

If a character has the _syntax type_ "constituent", it can be in a token. That's what "constituent" means. I absolutely think we should have characters not in this table, such as almost all of Unicode, available as constituents.

Once a token has been accumulated by the reader, it's parsed into whatever based on the _constituent traits_ (different from syntax type) of its characters. The nature of a token is pretty well defined in section 2.3, based on its characters' constituent traits. There are only three kinds of tokens - numbers, symbols, and consing dot - or it could be invalid, if it has incoherent syntax or characters with the "invalid" trait like Backspace. The syntax of tokens is defined in terms of the existing constituent traits. I don't think Clasp needs other kinds of tokens or to have different syntax for tokens, so we don't need new constituent traits.

kpoeck:
Ok, will adapt as you say in the last paragraph, and run regression tests and ansi-tests to validate.

Bike:
alright, seems fine. if you can fix that conflict i'll merge.

kpoeck:
I will write 1000 time on the board
- you shall not change the same file in 2 different pull requests
- you shall not change the same file in 2 different pull requests
....

kpoeck:
ok, no more merge conflict

PR #836, opened 2019-07-28
Fix write-string for unicode input

### LINKED ISSUE #835: Optimized write-to-string does not work for unicode strings
I believe Bike said that already on irc clasp
```lisp
;;; ansi-test write-to-string.1 simplified
(let ((unicode-string (make-array 4 :element-type 'character :initial-contents (mapcar #'code-char (list 40340 25579 40824 28331)))))
  (string=
   (with-output-to-string (s)(write unicode-string :stream s))
   (write-to-string unicode-string)))

(let ((unicode-string (make-array 4 :element-type 'character :initial-contents (mapcar #'code-char (list 40340 25579 40824 28331)))))
  (values
   (with-output-to-string (s)(write unicode-string :stream s))
   (write-to-string unicode-string)))

"\"鶔揫齸溫\""
"\"x\""
````

kpoeck:
* fixes #835
* passes regression tests
* passes ansi-tests write-to-string.1

### DIFF
--- a/include/clasp/core/lispStream.h
+++ b/include/clasp/core/lispStream.h
@@ -175,8 +175,9 @@ T_sp cl__make_synonym_stream(T_sp sym);
 T_sp cl__make_two_way_stream(T_sp in, T_sp out);
 
 T_sp cl__make_string_input_stream(String_sp strng, cl_index istart, T_sp iend);
-T_sp clasp_make_string_output_stream(cl_index line_length = 128, bool extended = false);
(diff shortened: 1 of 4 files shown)


Bike:
preprocessor constant names should be in allcaps, and we probably don't need the "CLASP_". otherwise seems fine.

unicode buffers aren't any _slower_ than base buffers, right? just bigger, and since this one is reused that doesn't really matter

kpoeck:
Changed what you requested

PR #839, opened 2019-08-03
ansi-tests: (setf (slot-value ..)) should ignore values of slot-missing

kpoeck:
Fixes test slot-missing-2-simplified and the corresponding ansi-test


Bike:
Given that `(setf slot-value)` should always return the value, we can probably just throw it at the end. That is, we shouldn't need the value from `(setf slot-value-using-class)` either. Looks nicer. Also it's what ECL does as of six years ago so I don't know why we don't. Maybe I missed it up.

kpoeck:
Ok, did as you said (and compared to ecl). When testing is done, will push an update

kpoeck:
pushed

PR #840, opened 2019-08-05
Fix equalp hash tables char as key

kpoeck:
fixes:
* equalp-hash-table-1


kpoeck:
also fixes ansi-test  equalp.32

PR #841, opened 2019-08-07
Flamegraphs on Linux generated with perf profiler

selwynsimsek:
This adds support for the generation of flamegraphs on Linux using the perf profiler, which has been in demand for a while now.

The resulting flamegraph for the benchmark
(loop for x below 100 do (compile-file "sys:kernel;lsp;foundation.lsp"))
is as deep as what is produced on macOS:
[macOS](https://kpoeck.github.io/out-20942.svg)
[linux](https://selwynsimsek.github.io/out-22766.svg)
The stack traces are generated using DWARF debug info.
The flamegraphs are generated just as in macOS:

cd src/profiler
./do-flame $PID

$FLAME_GRAPH_HOME must be set to the appropriate directory (as it should be for macOS)

It takes me 10 minutes and 500MB of perf output to produce one flamegraph - not sure if this is excessive


drmeister:
Thank you for this - it took a while to get around to merging it.

PR #842, opened 2019-08-09
Fix equalp for Str8Ns_O and StrWNs_O

kpoeck:
Fixes:
* equalp-2
* equalp-2a
* equalp-2b
* equalp-2c

By defining equalp for 
* Str8Ns_O
* StrWNs_O

Added tests for the change
Did run regression tests and complete ansi-tests with cst compiler

### DIFF
--- a/include/clasp/core/array.h
+++ b/include/clasp/core/array.h
@@ -1319,6 +1319,7 @@ namespace core {
     virtual clasp_elttype elttype() const { return clasp_aet_bc; };
   public:
     virtual bool equal(T_sp other) const final;
+    virtual bool equalp(T_sp other) const final;
   public:
     iterator begin() { return &(*this)[0]; };
     iterator end() { return &(*this)[this->length()]; };
@@ -1376,6 +1377,7 @@ namespace core {
     virtual clasp_elttype elttype() const { return clasp_aet_ch; };
   public:
     virtual bool equal(T_sp other) const final;
+    virtual bool equalp(T_sp other) const final;
   public:
     iterator begin() { return &(*this)[0]; };
     iterator end() { return &(*this)[this->length()]; };
--- a/src/core/array.cc
+++ b/src/core/array.cc
@@ -2261,6 +2261,17 @@ bool Str8Ns_O::equal(T_sp other) const {
   return false;
 };
 
+bool Str8Ns_O::equalp(T_sp other) const {
+  if (&*other==this) return true;
+  if (!other.generalp()) return false;
+  if (cl__stringp(other)) {
+    String_sp sother = gc::As_unsafe<String_sp>(other);
+    TEMPLATE_HALF_STRING_DISPATCHER(this,sother,template_string_equalp_bool,0,this->length(),0,sother->length());
+  } else {
+    return this-> MDArray_O::equalp(other);
+  }
+}
+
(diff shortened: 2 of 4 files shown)


Bike:
Please describe the problem and how you solved it. I'd really like you to do this whenever you file a PR.

I'm _guessing_ from the tests that it has to do with strings with fill pointers, but what about the regular array method isn't adequate?

kpoeck:
As I tried to explain by the tests that are failing, the problem is with equalp for Strings that are not simple. All the tests 
* equalp-2
* equalp-2a
* equalp-2b
* equalp-2c
in latest clasp (cst, no local changes):
```lisp
COMMON-LISP-USER> (equalp
       (make-array 12 :element-type 'character :initial-contents "123456789012" :fill-pointer 10)
       (make-array 15 :element-type 'character :initial-contents "123456789012345" :fill-pointer 10))

NIL
COMMON-LISP-USER> (equalp
       (make-array 15 :element-type 'base-char :initial-contents "123456789012345" :fill-pointer 10)
       (make-array 12 :element-type 'base-char :initial-contents "123456789012" :fill-pointer 10))

NIL
COMMON-LISP-USER> (equalp
       (make-array 3 :element-type 'base-char :initial-contents "abc" :fill-pointer 2)
       (vector #\a #\B))

NIL
COMMON-LISP-USER> (equalp
       (make-array 3 :element-type 'character :initial-contents "abc" :fill-pointer 2)
       (vector #\a #\B))

NIL
````

Looking at the relevant hierarchy of the implementation we see the following class hierarchy:
````
General_O
--Array_O
----MDArray_O
------ComplexVector_O
--------StrNs_O
----------Str8Ns_O
----------StrWNs_O
----AbstractSimpleVector_O
------SimpleString_O
--------template_SimpleVector<SimpleBaseString_O,claspChar,SimpleString_O> specialized_SimpleBaseString;
----------SimpleBaseString_O
--------template_SimpleVector<SimpleCharacterString_O,claspCharacter,SimpleString_O> specialized_SimpleCharacterString;
----------SimpleCharacterString_O
````
Implementations of equalp for arrays are in
template_Vector 
MDArray_O::equalp
AbstractSimpleVector_O::equalp
SimpleBaseString_O::equalp
SimpleCharacterString_O::equalp

Non-simple strings are Str8Ns_O & StrWNs_O (8 bit and wide strings)
Note that Simplestrings SimpleBaseString_O & SimpleCharacterString_O do have implementations for equalp.

The regular method that fires here is from template_Vector  that returns false since the dimensions do not match:
`if (this->_Dimensions[i] != other->_Dimensions[i]) return false;`

By analogy to SimpleBaseString_O::equal and SimpleCharacterString_O::equalp I therefor defined 
* StrWNs_O::equalp &
* Str8Ns_O::equalp

and the test pass

kpoeck:
Looking at the code and my own explanation, I wonder whether it would work just with defining a method for 
* StrNs_O &
* SimpleString_O

Instead of the 4 methods that exist after my patch, will verify this now

kpoeck:
nope, does not seem to be possible because of the templates

kpoeck:
@Bike is there anything more I can do here? I believed I described all what I should on 10th of August

Bike:
I get it now. Will merge later.

PR #845, opened 2019-09-02
Initial for for float-bits conversion functions for issue 583

kpoeck:
provide missing functions for float-features as requested in issue 583 by shinmera:
- core::single-float-bits (single-float to bits)
- core::single-float-from-unsigned-byte-32 (bits to single float)
- core::double-float-bits (double floats to bits)
- core::double-float-from-bits (bits to double floats)
- tests for all of the above


drmeister:
Looks good - I'm merging this. Thank you!

PR #849, opened 2019-09-13
Allow meta-dot for clasp c++ and h files (after make TAGS)

kpoeck:
* execute make TAGS (name taken from ecl)
* meta-dot in c++ and h files work, once TAGS file is known to emacs

### DIFF
--- a/makefile
+++ b/makefile
@@ -23,6 +23,10 @@ clean-modules:
 	find src/lisp/modules/ . -name '*.fasl' -exec rm {} \;
 	find src/lisp/modules/ . -name '*.o' -exec rm {} \;
 
+TAGS:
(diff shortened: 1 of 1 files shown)


Bike:
We don't actually like... use the makefile, do we? It's all waf.

kpoeck:
Well the makefile calls waf, so I often simply use make to recompile. But I can certainly look into wscript to integrate the same if you prefer that

Bike:
I'm not in charge of the build system and don't want to be. drmeister or cracauer should decide such things.

Anyway, I just tested it out and the tags table works fine, so no problem there

kpoeck:
assigning to cracauer and drmeister as Bike proposed

Bike:
Chris said it's fine in person so I'll just merge

Issue #850, opened 2019-09-14
Ironclad does not compile with cst


kpoeck:
There are 3 issues:
* one needs export GC_INITIAL_HEAP_SIZE=10G to avoid running out of memory
* ironclad changes the *readtable* but not eclectors readtable. Will try to teach eclector clasp readtable as agreed on #clasp. 
* Compiling ironclad cached-function-scope is called with nil and that breaks

kpoeck:
to work around the second issue, change the following in package.lisp
```lisp
(defun array-reader (stream subchar arg)
  (declare (ignore subchar))
  (let ((array-data (read stream nil stream nil))
        (array-element-type `(unsigned-byte ,arg)))
    ;; FIXME: need to make this work for multi-dimensional arrays
    `(make-array ,(length array-data) :element-type ',array-element-type
                :initial-contents ',array-data)))

    #-cst (set-dispatch-macro-character #\# #\L #'sharpL-reader)
    #+cst (eclector.readtable:set-dispatch-macro-character
     eclector.readtable:*readtable*
     #\# #\L #'sharpL-reader)
     
(defparameter *ironclad-readtable*
  #-cst
  (let ((readtable (copy-readtable nil)))
    (set-dispatch-macro-character #\# #\@ #'array-reader readtable)
    readtable)
  #+cst
  (progn
  (eclector.readtable:set-dispatch-macro-character
     eclector.readtable:*readtable*
     #\# #\@ #'array-reader)
     (copy-readtable nil)))
````

This is obviously no solution, but allows one to come to the last issue

kpoeck:
to work around the last issue:
```lisp
(defun get-dilocation (spi)
  (let ((lineno (core:source-pos-info-lineno spi))
        (col (core:source-pos-info-column spi))
        (inlined-at (core:source-pos-info-inlined-at spi)))
    (when (and *trap-zero-lineno* (zerop lineno))
      (format *error-output* "In get-dilocation lineno was zero! Setting to ~d~%"
              (setf lineno 666666)))
    (if inlined-at
        (let ((fs (core:source-pos-info-function-scope spi)))
          (when fs
            (llvm-sys:get-dilocation
             *llvm-context* lineno col
             (cached-function-scope fs)
             (get-dilocation inlined-at))))
        (llvm-sys:get-dilocation *llvm-context* lineno col *dbg-current-scope*))))

(defun dbg-set-irbuilder-source-location (irbuilder spi)
[...]

PR #852, opened 2019-09-18
[Wip] Define core:defconstant-equalp

kpoeck:
to allow `(core:defconstant-equalp +foo+ (make-array 3 :initial-contents (list 1 2 3)))`

### DIFF
--- a/src/lisp/kernel/lsp/evalmacros.lsp
+++ b/src/lisp/kernel/lsp/evalmacros.lsp
@@ -101,7 +101,10 @@ VARIABLE doc and can be retrieved by (DOCUMENTATION 'SYMBOL 'VARIABLE)."
 (defmacro defconstant-equal (var form &optional doc-string)
   `(defconstant-eqx ,var ,form equal ,doc-string))
 
-(export '(defconstant-equal))
+(defmacro defconstant-equalp (var form &optional doc-string)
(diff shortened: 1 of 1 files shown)


Bike:
I don't mind adding this, but where it would it be used?

kpoeck:
For defconstant of arrays as i put in the example, they are not equal but only equalp ( apart from strings and bitvectors)

kpoeck:
like in the issue #850: I had to put (using something unexported):
```lisp
(core::defconstant-eqx round-constants
  (make-array 10 :initial-contents (list #x01000000 #x02000000 #x04000000 #x08000000 #x10000000
                                         #x20000000 #x40000000 #x80000000 #x1B000000 #x36000000))
  equalp)
````

Bike:
Put where? Ironclad isn't going to use anything in our CORE package. Our tests?

kpoeck:
Core:defconstant-equal does not work for arrays, but i said that already.

I consider it bad style if i need to access a core::whatever to define a constant array. I cannot use defconstan either

Bike:
That's not what I'm asking. I'm asking what code you are writing that needs to define equalp constants, that's all. WHERE are you defining these constant arrays. In the issue you linked it's just a test case for a bug, and the constantness probably doesn't even matter.

Issue #853, opened 2019-09-19
Regression tests: compiler macro for assoc gives SEGMENTATION-VIOLATION on incorrect input


kpoeck:
Test assoc-2-error
with compiler macro
```lisp
COMMON-LISP-USER>> (let ()(assoc 3 (list (list 1 3) 2)))
Debugger received error of type: SEGMENTATION-VIOLATION
Segmentation fault. Attempted to access resticted memory address #x5.
This is due either to a problem in foreign code (e.g., C++), or a bug in Clasp itself.
Error flushed.
````
but without compiler macro
```lisp
COMMON-LISP-USER>> (let ()(locally (declare (notinline assoc))
                      (assoc 3 (list (list 1 3) 2))))
Debugger received error of type: TYPE-ERROR
2 is not of type LIST.
Error flushed.
````

kpoeck:
Backtrace is completely unhelpful

kpoeck:
The problem is obviously the (the cons ,%elt) in expand-assoc 
see https://github.com/clasp-developers/clasp/blob/dev/src/lisp/kernel/cmp/opt-cons.lsp#L100
Obviously it is proven that %elt is a cons, so the declaration more says that it should be a cons.

Cons_O::assoc does test whether the current element is really a cons

I wonder whether the regression tests should do 
(DECLAIM (OPTIMIZE (SAFETY 3) (SPEED 0)))
or whatever is needed that (car (the cons foo)) does not expand to unsafe code

@Bike, your opinion?

kpoeck:
(let () (sublis '((a . 1) . bad) (list 'a 'b 'c 'd))) also gives a SEGMENTATION-VIOLATION.

Unfortunatey this cannot be cured even by (declaim (optimize (safety 3)(speed 0))) unless sublis is recompiled.

kpoeck:
same for 
```lisp
 (declaim (optimize (speed 0)(safety 3)))

((CLASP-CLEAVIR::SAVE-REGISTER-ARGS . T) (CORE::INSERT-ARRAY-BOUNDS-CHECKS . T) (CLASP-CLEAVIR::DO-TYPE-INFERENCE) (CLASP-CLEAVIR::DO-DX-ANALYSIS) (CLEAVIR-ESCAPE:TRUST-DYNAMIC-EXTENT) (CLEAVIR-KILDALL-TYPE-INFERENCE:INSERT-TYPE-CHECKS . T))
COMMON-LISP-USER>> (every
[...]

PR #854, opened 2019-09-20
Run regression-tests with safety set to high, adapt expected errors!

kpoeck:
* w/o (safety 3) some tests segfault on incorrect arguments instead of raising an type error (see #853)


kpoeck:
* solved the merge conflict
* added tests for the segementation violation problem (now solved)

kpoeck:
* and added more equalp tests

PR #856, opened 2019-09-22
Undo some potentially wrong declarations that lead to segmentation faults

### LINKED ISSUE #853: Regression tests: compiler macro for assoc gives SEGMENTATION-VIOLATION on incorrect input

kpoeck:
Fixes #853 and #854 by dropping potentially wrong declarations and for vector-merge-sort actually test that the argument is a vector

Did full rebuild, run regression-test and ansi-tests to validate the changes and added tests to the regression tests.

With the changes ansi-tests do run again w/o segmentation violations.

### DIFF
--- a/src/lisp/kernel/cmp/opt-cons.lsp
+++ b/src/lisp/kernel/cmp/opt-cons.lsp
@@ -33,7 +33,7 @@
 (defmacro do-in-list ((%elt %sublist list &rest output) &body body)
   `(do* ((,%sublist ,list (cdr ,%sublist)))
         ((null ,%sublist) ,@output)
-     (let ((,%elt (car (the cons ,%sublist))))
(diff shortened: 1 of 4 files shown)


kpoeck:
solved otherwise

PR #862, opened 2019-10-04
Fix encodings in clasp (not fully ported from ecl)

kpoeck:
* ext__decoding_error was calling itself in a recursion with wrong parameters (but should call into lisp). There is no need to be callable from lisp
* ext__decoding_error was calling itself in a recursion with wrong parameters but should call into lisp). There is no need to be callable from lisp
* core__load_source did ignore the externalFormat and always used default
* encoding and decoding conditions where forgotten
* moved the encoding stuff to new file encodings.lisp to be added in cclasp
* ext:load-encoding did nothing usefull, so disabled it 
* changed ext:make-encoding to use encoding/decoding tables generated from ecl, added aliases for the encoding as in ecl.
* generation is done in ecl executing create-encodings-from-ecl
* added tests
* did rebuild with distclean
* did successfuly run regression tests
* did successfully run ansi-test


Bike:
Do we need those commented out definitions with `#+(or)`? Does ECL have to be used if we want to generate another encoding, like what are we missing in Clasp to do it if the code's there? Why is there a comment in the tests saying Latin-2 doesn't work?

kpoeck:
We don't need to the #+(or) so I delete them. 
The latin-2 comment is no longer true, so let me delete that.
create-encodings-from-ecl needs to be run from ecl, but the  generator could be adapted to use other sources

kpoeck:
Should have done what you proposed

PR #864, opened 2019-10-11
Remove useless T_mv

kpoeck:
* dropped T_mv where only 1 value was returned and clhs does not ask for multiple values
* tighten the type for cl___PLUS_
* tighten the type for cl___TIMES_
* tighten the type for cl___MINUS_
* tighten the type for cl___TIMES_
* tighten the type for cl___DIVIDE_
* tighten the type for cl__atan
* tighten the type for cl__rational
* tighten the type for cl__read_delimited_list

Did run regression and ansi-tests to validate (with sucess)


Bike:
Fixing the conflicts locally, will merge in a bit.

PR #865, opened 2019-10-12
Update regression tests for bitarrays

kpoeck:
* Test #0A0 and #0A1 in one place instead of every single test
* update expected failures with  less failure due to latest bit-array refactoring (HELL-MDBIT-5A)


kpoeck:
* and (bit-and #0A0 #0A0) fails, since #(0) is not of type SIMPLE-BIT-VECTOR

PR #872, opened 2019-10-19
Define ComplexVector_O::__write__, so that the fill pointer is respected

### LINKED ISSUE #870: Regression PRINT-5 (clasp-regression-tests)
```lisp
COMMON-LISP-USER> (string= "#()"
               (WRITE-TO-STRING
                (MAKE-ARRAY '(4)
                            :INITIAL-CONTENTS
                            '(0 1 2 3)
                            :ELEMENT-TYPE
                            '(UNSIGNED-BYTE 2)
                            :FILL-POINTER
                            0)
                :READABLY NIL :ARRAY T :PRETTY NIL))

NIL
COMMON-LISP-USER> (WRITE-TO-STRING
                (MAKE-ARRAY '(4)
                            :INITIAL-CONTENTS
                            '(0 1 2 3)
                            :ELEMENT-TYPE
                            '(UNSIGNED-BYTE 2)
                            :FILL-POINTER
                            0)
                :READABLY NIL :ARRAY T :PRETTY NIL)

"#(0 1 2 3)"
````

kpoeck:
* Fixes #870 
* added another test for the problem

### DIFF
--- a/include/clasp/core/array.h
+++ b/include/clasp/core/array.h
@@ -398,6 +398,8 @@ class ComplexVector_O : public MDArray_O {
                   Array_sp data,
                   bool displacedToP,
                   Fixnum_sp displacedIndexOffset) : MDArray_O(Rank1(),dimension,fillPointer,data,displacedToP,displacedIndexOffset) {};
+    public:
+    virtual void __write__(T_sp strm) const;
 };
 };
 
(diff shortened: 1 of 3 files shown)


Bike:
Right idea but doesn't account for `*print-length*` or element types. I'm extending this now

Issue #875, opened 2019-11-03
Regression: PRINT.RATIOS.RANDOM.root-cause fails


kpoeck:
```lisp
COMMON-LISP-USER> (let ((*PRINT-BASE* 22)
                (*print-radix* t))
            (write-to-string -59990859179/64657108615))

"#22r-1212B/13K5KA78B"
````
the correct result would be 
"#22r-1212B0C39/13K5KA78B".

kpoeck:
The error seem to be starting  here:
https://github.com/clasp-developers/clasp/blob/dev/src/core/numberToString.cc#L63
where the buffer should be extended
but in
https://github.com/clasp-developers/clasp/blob/dev/include/clasp/core/string.h#L198
probably the comparison os wrong and should be (min > (min > this->_ArrayTotalSize))

kpoeck:
fixed

PR #876, opened 2019-11-03
Fix issue 875

kpoeck:
Fix issue #875 
* In fillPointerSet verify that the vector is big enough (please review whether this is the right place to do that)
* In ensureSpaceAfterFillPointer fix comparison
* in core__bignum_to_string use the specific pointer Str8Ns_sp or StrWNs_sp


Bike:
seems like the right place to check the fill pointer if we're gonna.

i'm curious why the buffer->buffer8 etc thing is necessary

PR #896, opened 2019-12-15
Apply translate-logical-pathname to passed pathanme in bundle-to-library-file-name

kpoeck:
* with that the regression-tests run again after the fasl-directory change


kpoeck:
Fasl directory change seems to be retired

PR #899, opened 2019-12-27
Adapt calls parameter to eclector

kpoeck:
reorder parameters so that they actualy fir eclectors definition


drmeister:
I checked all of the eclector-readtable function call arguments and kpoeck is correct - Clasp needs to incorporate this pull request to agree with the eclector function call argument orders.

PR #901, opened 2020-01-05
Fix issue 900

### LINKED ISSUE #900: llvm@9: segmentation violation calling fill-pointer with #0A2
```lisp
COMMON-LISP-USER> (fill-pointer (make-array nil :initial-element 2 :fill-pointer nil))

Condition of type: SEGMENTATION-VIOLATION
Segmentation fault. Attempted to access resticted memory address #x0.

Could not format ~&Available restarts:~&(use :r1 to invoke restart 1)~2% NIL1. (RESTART-TOPLEVEL) Go back to Top-Level REPL.

Broken at frame[0] NIL. In: #<PROCESS TOP-LEVEL @0x11127eb59>.
;;; Warning: compiled-function-file expected a function as argument - but it got NIL - there may not be any backtrace available
````

kpoeck:
* fixes #900 
* check for nil dimension array in 
* added test
* did run regression test with no new failures, but problem fixed
* did run ansi-test, test that gave segmentation fault works now

### DIFF
--- a/src/core/array.cc
+++ b/src/core/array.cc
@@ -674,6 +674,9 @@ struct RecursivePrint {
   }
 
   bool advanceIndices(int level) {
+    // if this is a nil sized array, indices is empty, so already return false, otherwise we get EXC_BAD_ACCESS  
+    unlikely_if (this->indices.empty())
+        return false;
     this->indices[level]++;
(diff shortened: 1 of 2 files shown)


drmeister:
Looks good - thank you!

PR #904, opened 2020-01-06
Set initial stacksize again to 16 Mb

kpoeck:
* put code to set rl.rlim_cur  back in (no need to fiddle with rl.rlim_max), only if current limit is smaller than desired limit
* change wscript so that starting from mojave '-Wl,-stack_size,0x1000000' is not passed to the link options (if done, the setrlimit operation from above fails, see irc log for an explanation)
* fix spelling of def macos_version


drmeister:
Good.

PR #916, opened 2020-02-01
Fix compile-file and compile-file-parallel to respect fasl extensions when appropriate

kpoeck:
* regression test did not work, since compile-file (parallel) coudn't handle logical pathname
* cross-compiling sbcl did not not work, since the supplied fasl extension was ignored. Sbcl will later remain the fasl-extension for a standard one, so that it can be loaded 
* added regression tests to verify


kpoeck:
and there is a merge conflict, heavanly

kpoeck:
There should also be a conflict  for  src/lisp/kernel/cmp/compile-file-parallel.lsp, don't understand this

kpoeck:
Will retry with newest sources

Issue #885, opened 2019-12-05
Alexandria test 


kpoeck:
clasp --eval "(require :asdf)" --load "~/quicklisp/setup.lisp" 
```lisp
Starting cclasp-boehm-0.4.2-1662-g8d0a04160-cst ... loading image...
Top level in: #<PROCESS TOP-LEVEL @0x113013279>.
COMMON-LISP-USER> (ql:quickload "alexandria-tests" :verbose t)
To load "alexandria-tests":
  Load 1 ASDF system:
    alexandria-tests
; Loading "alexandria-tests"

("alexandria-tests")
COMMON-LISP-USER> (alexandria-tests::run-tests :compiled t)

Doing 227 pending tests of 227 tests total.
 ALEXANDRIA-TESTS::COPY-ARRAY.1 ALEXANDRIA-TESTS::COPY-ARRAY.2
 ALEXANDRIA-TESTS::COPY-ARRAY.3 ALEXANDRIA-TESTS::COPY-ARRAY.4
 ALEXANDRIA-TESTS::ARRAY-INDEX.1 ALEXANDRIA-TESTS::UNWIND-PROTECT-CASE.1
 ALEXANDRIA-TESTS::UNWIND-PROTECT-CASE.2
 ALEXANDRIA-TESTS::UNWIND-PROTECT-CASE.3
 ALEXANDRIA-TESTS::UNWIND-PROTECT-CASE.4
 ALEXANDRIA-TESTS::UNWIND-PROTECT-CASE.5 ALEXANDRIA-TESTS::SWITCH.1
 ALEXANDRIA-TESTS::SWITCH.2 ALEXANDRIA-TESTS::ESWITCH.1
 ALEXANDRIA-TESTS::ESWITCH.2 ALEXANDRIA-TESTS::CSWITCH.1
 ALEXANDRIA-TESTS::CSWITCH.2 ALEXANDRIA-TESTS::MULTIPLE-VALUE-PROG2.1
 ALEXANDRIA-TESTS::NTH-VALUE-OR.1 ALEXANDRIA-TESTS::WHICHEVER.1
 ALEXANDRIA-TESTS::WHICHEVER.2 ALEXANDRIA-TESTS::XOR.1
 ALEXANDRIA-TESTS::XOR.2 ALEXANDRIA-TESTS::XOR.3
 ALEXANDRIA-TESTS::DEFINE-CONSTANT.1 ALEXANDRIA-TESTS::DEFINE-CONSTANT.2
[...]

PR #917, opened 2020-02-02
Update CST (issue 885) and get asdf version 3.3.3.5

kpoeck:
* update cst to version solving issue #885 and from updated git repository
* get asdf from version 3.3.3.5 with all clasp fixes (and not via commit id)


drmeister:
Great!

PR #918, opened 2020-02-02
Respect passed extension when possible (for sbcl cross-compile) and silence messages unless :verbose is true

kpoeck:
* regression tests did not work, since compile-file (parallel) coudn't handle logical pathname
* cross-compiling sbcl did not not work, since the supplied fasl extension was ignored. Sbcl will later remain the fasl-extension for a standard one, so that it can be loaded
* most ansi-tests for compile-file failed
* added regression tests to verify
* did run regression-test
* did run ansi-tests


kpoeck:
tested with
* build from scratch
* regression test
* ansi-tests

drmeister:
I'm merging this and then I'll change this call...

(core:write-faso (translate-logical-pathname output-path) (mapcar #'cdr sorted-object-files)))))

So that the (translate-logical-pathname output-path) is invoked inside of core:write-faso.

PR #921, opened 2020-02-02
Feature small defstruct fixes

kpoeck:
Some fixes from ansi-test and sbcl crosscompiling and assorted tests.
* (defstruct foo-0a (bar 42 :read-only nil)) ignored the nil after read-only
* Fix for "Mutable slot ~a cannot override read-only included slot." contributed By Bike in irc
* `(DEFSTRUCT (STRUCT-TEST-45 (:TYPE LIST) (:INITIAL-OFFSET 2)) A45 B45)` and `(DEFSTRUCT
              (STRUCT-TEST-47 (:TYPE LIST) (:INITIAL-OFFSET 3)
               (:INCLUDE STRUCT-TEST-45))
              C47
              D47)`gave an incorrect error about slot nil repeating
* defstruct ..... (:TYPE LIST) :NAMED) generated wrong predicate function (test for 1 slot too much)
* some errors by the tests are still open, I believe it is not enough to look at the direct include, but need to look up the whole chain of includes
  * INCLUDE-LEVEL-2 
  * SBCL-CROSS-COMPILE-1
  * SBCL-CROSS-COMPILE-4

### DIFF
--- a/src/lisp/kernel/lsp/defstruct.lsp
+++ b/src/lisp/kernel/lsp/defstruct.lsp
@@ -136,7 +136,8 @@
                        (if seen-read-only
                            (error-defstruct-slot-syntax slot-description)
                            (setq seen-read-only t
-                                 read-only t)))
+                                 ;;; treat :read-only nil correctly
+                                 read-only (cadr os))))
                       (otherwise
                        (error-defstruct-slot-syntax slot-description)))))
                  (t (error-defstruct-slot-syntax slot-description))))
@@ -184,8 +185,9 @@
     (let ((old-reader (getf old-plist :reader))
           (old-accessor (getf old-plist :accessor)))
       (when (and accessor old-reader)
-        (error "Mutable slot ~a cannot override read-only included slot."
-               slot-name))
+        #+(or)(error "Mutable slot ~a cannot override read-only included slot."
+                     slot-name)
+        (setq reader accessor accessor nil))
       `(,slot-name
         ;; We always have an initarg for any non-:NAMED slot.
         :initarg ,initarg
@@ -255,9 +257,11 @@
           (error "Cannot override nonexistent slot ~a" slot-name))))
     ;; Signal an error for any duplicate slot.
     (dolist (new-slotd new-slotds)
-      (let ((slot-name (first new-slotd)))
-        (when (member slot-name old-slotds :key #'first)
-          (error "Duplicate slot ~a" slot-name))))
+      ;;; check for "dummy" slotdescriptions coming from :initial-offset
+      (unless (null new-slotd)
+        (let ((slot-name (first new-slotd)))
+          (when (member slot-name old-slotds :key #'first)
+            (error "Duplicate slot ~a" slot-name)))))
     ;; Done.
     (append old-slotds new-slotds)))
 
@@ -512,6 +516,13 @@
        `(set-documentation ',structure-name 'structure
                            ',(second option))))))
 
+(defun named-slot-description-p (structure-name slot-description)
+  (let* ((nsd (named-slot-description structure-name))
+         (sym1 (first slot-description))
+         (sym2 (first nsd)))
+    (and (symbolp sym1)(symbolp sym2)(string= (symbol-name sym1)(symbol-name sym2))
+         (equalp (rest nsd) (rest slot-description)))))
+                                 
 (defun defstruct-list-option-expander
     (structure-name included-size slot-descriptions)
   (lambda (option)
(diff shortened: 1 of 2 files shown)


Bike:
Can you explain this `named-slot-description-p` stuff? My intent was that we shouldn't need to do any test like that.

kpoeck:
Problem is with
`(DEFSTRUCT (STRUCT-TEST-38 (:TYPE LIST) :NAMED) A38 B38 C38)`
generating a wrong predicate.
```lisp
(macroexpand '(DEFSTRUCT (STRUCT-TEST-38 (:TYPE LIST) :NAMED) A38 B38 C38))
->
(DEFUN STRUCT-TEST-38-P (CORE::OBJECT)
   (AND (CONSP CORE::OBJECT)
        (CONSP (SETF CORE::OBJECT (CDR CORE::OBJECT)))
        (CONSP (SETF CORE::OBJECT (CDR CORE::OBJECT)))
        (CONSP (SETF CORE::OBJECT (CDR CORE::OBJECT)))
        (CONSP (SETF CORE::OBJECT (CDR CORE::OBJECT)))
        (EQ (NTH 0 CORE::OBJECT) 'STRUCT-TEST-38)))
...

this tests 4 elements in the list, but there are only 3, so
```lisp
(STRUCT-TEST-38-P (make-struct-test-38))
nil
````

**Comment by kpoeck** [comment]
the problem is in  defstruct-list-option-expander or the parameter slot-descriptions as passed:
```lisp
COMMON-LISP-USER> (trace core::defstruct-list-option-expander)

(CORE::DEFSTRUCT-LIST-OPTION-EXPANDER)
COMMON-LISP-USER> (DEFSTRUCT (STRUCT-TEST-38 (:TYPE LIST) :NAMED) A38 B38 C38)
1> (CORE::DEFSTRUCT-LIST-OPTION-EXPANDER STRUCT-TEST-38 0 ((#:STRUCT-TEST-38
                                                            :INITFORM
                                                            'STRUCT-TEST-38)
                                                           (A38 :ACCESSOR
                                                            STRUCT-TEST-38-A38
                                                            :INITARG :A38)
                                                           (B38 :ACCESSOR
                                                            STRUCT-TEST-38-B38
                                                            :INITARG :B38)
                                                           (C38 :ACCESSOR
                                                            STRUCT-TEST-38-C38
                                                            :INITARG :C38)))
<1 (CORE::DEFSTRUCT-LIST-OPTION-EXPANDER #<FUNCTION LAMBDA>)
````
`(#:STRUCT-TEST-38 :INITFORM 'STRUCT-TEST-38)`seem to correspond to the :named part

**Comment by kpoeck** [comment]
and tracing NAMED-SLOT-DESCRIPTION 
```lisp
1> (CORE::NAMED-SLOT-DESCRIPTION STRUCT-TEST-38)
<1 (CORE::NAMED-SLOT-DESCRIPTION (#:STRUCT-TEST-38 :INITFORM 'STRUCT-TEST-38))
````

**Comment by Bike** [comment]
I'm seeing some pretty deep screwiness here. The predicate is definitely wrong but so is everything else. For example, if I (make-struct-test-38) i get (NIL NIL NIL NIL) but I ought to be getting (STRUCT-TEST-38 NIL NIL NIL). And it looks like I should be from the `(DEFUN MAKE-STRUCT-TEST-38 ...)` in the macroexpansion.

**Comment by Bike** [comment]
Alright so here's the predicate definition from the code

```lisp
`(defun ,(second option) (object)
   (and
    (consp object)
    ,@(let (forms)
        (dotimes (i (length slot-descriptions) forms)
          (push '(consp (setf object (cdr object))) forms)))
    ;; FIXME: inefficient
    (eq (nth ,(+ (third option) included-size) object)
        ',structure-name)))
```

Now to match the predicate we want two basic things: It should be a list of the correct length (at least) and the `:NAMED` pseudo-slot should be the right symbol. I managed to hork this up. The first part, the `dotimes`, checks that the length is correct without using `length` since we don't want to error if the list is dotted. It looks dumb but ought to work. But the second part that checks the pseudo-slot is wrong because it works with the `object` variable, which has by this point been set to `NIL` or something.

If we just use a different variable for the length check it ought to work. Or there could be a nicer way to do this since this is kind of lengthy and weird. But we shouldn't need to look for the named slot.

Bike:
```lisp
(defun list-of-length-at-least (list n)
  (dotimes (i n)
    (unless (consp list) (return-from list-of-length-at-least nil))
    (setf list (cdr list)))
  (listp list))

`(defun ,(second option) (object)
   (and (list-of-length-at-least object ,(length slot-descriptions))
        (eq (nth ,(+ (third option) included-size) object)
            ',structure-name)))
```

something like this?

kpoeck:
Looks good (you adressed the 2 problems). When I am back home after work, will run regression-test and ansi-test for defstructs to validate and given success update the pr

kpoeck:
applied what you said

Bike:
OK looks good. Might examine the mutable slot thing more later but whatever. I'd merge this now but drmeister asked me to hold off on pushing to dev today.

Bike:
also about the redefinition tests like i mentioned in irc - you could just save (fdefinition 'name) beforehand and see if it's EQ to the version after the new defstruct evaluation

kpoeck:
I take that proposal with the (fdefinition 'name), looks better than what I tried before

kpoeck:
Did as requested

PR #922, opened 2020-02-02
More tests

kpoeck:
regression-tests for a problem fixed by Bike


Bike:
Should probably use `locally` instead of `let ()` but whatevs

PR #924, opened 2020-02-11
[WIP] Further Fixes to include handling in defstruct

kpoeck:
* fix the remaining problems with include defstruct definitions from sbcl cross-compiling
   * generated accessor/readers even if the immediate include has now accessor/readers (include-level-2b)
   * check for duplicate definitions along the whole include chain, not only in the parent (include-level-3)
* new test struct-test-56 to check for regression introduced in a previous vesrion of the fix
* only known problem test  sbcl-cross-compile-4, since we don't detect an implicit read-only (code generated is ok, but error is silently corrected)

### DIFF
--- a/src/lisp/kernel/lsp/defstruct.lsp
+++ b/src/lisp/kernel/lsp/defstruct.lsp
@@ -53,6 +53,16 @@
   (put-sysprop name 'structure-constructor constructor))
 (defun names-structure-p (name)
   (structure-type name))
+(defun structure-include (name)
+  (get-sysprop name 'include))
+(defun (setf structure-include) (include name)
+  (put-sysprop name 'include include))
+
+(defun all-includes (struct-name)
+  (let ((include (structure-include struct-name)))
+    (if include
+        (cons include (all-includes include))
+        nil)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;
@@ -177,7 +187,7 @@
           `(,slot-name ,initform :read-only ,read-only :type ,type)))))
 
 ;;; Apply an :INCLUDE slot override.
-(defun override-slotd (slot-name over-plist old-plist)
+(defun override-slotd (slot-name over-plist old-plist include)
   (destructuring-bind (&key (initform nil initformp)
                          (type t typep) initarg
                          reader accessor)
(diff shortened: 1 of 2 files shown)


Bike:
i haaaaate defstruct

I wanted the code to work without needing to store include information like your code does. Like effective slots in normal CLOS. The `structure-slot-descriptions` are supposed to be complete. Right now they are not because they don't include older accessors. Probably they could be made complete by having more `:accessor` fields. Like, right now:

```lisp
CL-USER> (defstruct foo foo1 foo2)
FOO
CL-USER> (defstruct (bar (:include foo)) bar1 bar2)
BAR
CL-USER> (core::structure-slot-descriptions 'foo)
((FOO1 :ACCESSOR FOO-FOO1 :INITARG :FOO1)
 (FOO2 :ACCESSOR FOO-FOO2 :INITARG :FOO2))
T
CL-USER> (core::structure-slot-descriptions 'bar)
((FOO1 :INITARG :FOO1 :ACCESSOR BAR-FOO1)
 (FOO2 :INITARG :FOO2 :ACCESSOR BAR-FOO2)
 (BAR1 :ACCESSOR BAR-BAR1 :INITARG :BAR1)
 (BAR2 :ACCESSOR BAR-BAR2 :INITARG :BAR2))
T
```
But we could have
```lisp
CL-USER> (core::structure-slot-descriptions 'bar)
((FOO1 :INITARG :FOO1 :ACCESSOR BAR-FOO1 :ACCESSOR FOO-FOO1)
 (FOO2 :INITARG :FOO2 :ACCESSOR BAR-FOO2 :ACCESSOR FOO-FOO2)
 (BAR1 :ACCESSOR BAR-BAR1 :INITARG :BAR1)
 (BAR2 :ACCESSOR BAR-BAR2 :INITARG :BAR2))
T
```

and then have the redundant name checker look at all of them.

or it might be easier to have `:accessors (bar-foo1 foo-foo1)`, roughly analogous to `clos:slot-definition-writers`.

kpoeck:
Labled it WIP since we still discuss the design (so that nobody merges it by accident)

Bike:
just switching to `:readers` or whatever doesn't work since we need to know which one to actually have a definition for

I don't have a good design for this. I want a good design for this. I want to never have problems with defstruct again, or more realistically, when we find some bug or need to add some new features to it I want to not have to rewrite everything entirely. That's what I tried to do here and I failed. If you have a good idea for an actual organization - not just a fix, I mean - I'm all ears. Otherwise I'll just think harder.

PR #928, opened 2020-02-23
Small fixes for defstruct so that describe works better

kpoeck:
Small fixes in defstruct, so that describe/inspect works better:
* Fixes a typo in unparse-slot-description (`:read-only` instead of `:reader`
* also store the documentation for normal structs in `defstruct-class-option-expander`, not only for  type `list`and `vector`. Was already stored in the corresponding class
* indented `defstruct-class-option-expander`with emacs

### DIFF
--- a/src/lisp/kernel/lsp/defstruct.lsp
+++ b/src/lisp/kernel/lsp/defstruct.lsp
@@ -170,7 +170,7 @@
 (defun unparse-slot-description (list)
   (let ((slot-name (car list)) (plist (cdr list)) (default (list nil)))
     (let ((initform (getf plist :initform default))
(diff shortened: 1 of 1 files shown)


Bike:
the `set-documentation` call will go through `ext:annotate`, right?

kpoeck:
The 2 prs are complementary, and can be treated individually. This one has nothing to do with https://github.com/clasp-developers/clasp/pull/924.

And yes set-documentation still goes through ext:annotate, I just wanted to fix that structs of type list and vector store their documentation, but normal structs don't.

Bike:
Alright sure why not.

PR #929, opened 2020-02-23
Feature overhall describe

kpoeck:
* drop `si::get-documentation` and use `documentation`
* show lambda list when possible
* show source file when possible
* make compiled-function-file more robust
* describe structures more like sbcl
* new regressions test for all changed (documentation and describe)
* tested with
  * regression tests
  * ansi tests


drmeister:
Very nice - thank you!

PR #931, opened 2020-02-29
Silence compile-file verbose messages when verbose = nil

kpoeck:
* General messages about compilation should be controlled by verbose not by print. Print is supposed to give information for every compiled form.
* When :verbose nil is passed, it should not be overwritten by  *compile-verbose*. *compile-verbose* should be the default value for verbose
* Test already exists in the regression-test `COMPILE-FILE.1.simplified`


drmeister:
Looks great - thank you!

PR #934, opened 2020-03-07
Don't shoot myself in the knee redefining #'(Setf aref)

kpoeck:
* previously every (Setf (aref array x y) value) would do nothing after loading contol01.lisp


Bike:
Hang on, do you mean that one of those `setf fdefinition` forms actually is altering the fdefinition of `(setf aref)`?

PR #940, opened 2020-03-12
[Wip] Return sensible error in integer-decode-float when called with a infinity value

kpoeck:
* issue encountered while cross-compiling sbcl (and the origin is still to be clarified)
* return sensible error instead of sigfpe as before
* added test to regression-suite

### DIFF
--- a/src/core/num_co.cc
+++ b/src/core/num_co.cc
@@ -1109,6 +1109,8 @@ CL_LAMBDA(x);
 CL_DECLARE();
 CL_DOCSTRING("integer_decode_float");
 CL_DEFUN Real_mv cl__integer_decode_float(Float_sp x) {
+  unlikely_if (clasp_float_infinity_p(x))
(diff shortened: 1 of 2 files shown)


Bike:
Preferably we should figure out how to transmute a SIGFPE into a floating point error, because otherwise we'd have to put checks like this on every floating point operation. I thought I had it doing that already, but evidently not. Either the error isn't one of the ones we handle, or the traps are just totally messed up. Code is here: https://github.com/clasp-developers/clasp/blob/dev/src/gctools/interrupt.cc#L256-L271

kpoeck:
I hope in #955 I have this fixed as you intended

Bike:
yes. is this PR dead, then? incorporating the tests would be good, but #961 precludes that.

kpoeck:
As you said, this pr is dead. I have stored the tests locally, but as you said, we can't add them right now. I wonder whether we need something like  `(ext:issue-fixed-p :961)`

PR #941, opened 2020-03-12
Use :test 'eq for weak hast-tables as by recent change

kpoeck:
* weak hash-tables need to use :test 'eq (for the moment), default is eql


Bike:
Weak key hash tables seem too broken to use right now anyway, so maybe the tests aren't so important. But I guess we can use them later.

PR #947, opened 2020-03-21
[Wip] Fix hash-table-p for weak-hash-tables

### LINKED ISSUE #946: (hash-table-p (make-hash-table :test #'eq :weakness :key)) -> nil
Expected would be t.

Looking at the class hierarchy:
```lisp
(clos:class-direct-subclasses (first (clos:class-direct-superclasses (find-class 'core:weak-key-hash-table))))
(#<The BUILT-IN-CLASS HASH-TABLE>
 #<The BUILT-IN-CLASS CORE:WEAK-KEY-HASH-TABLE>)
````

perhaps the definition in https://github.com/clasp-developers/clasp/blob/dev/src/lisp/kernel/cleavir/inline.lisp#L182

should be changed 
from `hash-table-p hash-table`
to `hash-table-p  core:hash-table-base`

kpoeck:
* Fixes #946 
* Regression-test

### DIFF
--- a/src/lisp/kernel/cleavir/inline.lisp
+++ b/src/lisp/kernel/cleavir/inline.lisp
@@ -179,7 +179,7 @@
     core:double-float-p double-float
     floatp float
     functionp function
-    hash-table-p hash-table
+    hash-table-p core:hash-table-base
     integerp integer
     listp list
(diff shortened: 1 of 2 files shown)


kpoeck:
replaced by #1023

PR #955, opened 2020-03-28
Actually use handle_bus

kpoeck:
* if handle_signal_now is called -as it was before- , than ext::_sym_bus_error will be called with no arguments, which will crash, since ext:bus-error expects one argument
* Same with SIGSEGV in `ADD_SIGNAL( SIGSEGV, "SIGSEGV", ext::_sym_segmentation_violation);`
* `core::_sym_wait_for_all_processes` is not defined
* fix that `handle_fpe`is really called
* in `handle_fpe` handle all values for `info->si_code`, `FPE_NOOP`only on darwin

### DIFF
--- a/src/gctools/interrupt.cc
+++ b/src/gctools/interrupt.cc
@@ -255,15 +255,22 @@ int global_pollTicksGC = INITIAL_GLOBAL_POLL_TICKS_PER_CLEANUP;
 
 void handle_fpe(int signo, siginfo_t* info, void* context) {
   (void)context; // unused
-  init_float_traps(); // WHY
+  // printf("Enter handle_fpe Signo: %d Errno:%d Code:%d\n", (info->si_signo), (info->si_errno), (info->si_code));
+  // init_float_traps(); // WHY
   // TODO: Get operation and operands when possible.
   // Probably off the call stack.
   switch (info->si_code) {
+    #ifdef _TARGET_OS_DARWIN
+  case FPE_NOOP:   NO_INITIALIZERS_ERROR(cl::_sym_arithmeticError);
+    #endif
+  case FPE_INTDIV: NO_INITIALIZERS_ERROR(cl::_sym_divisionByZero);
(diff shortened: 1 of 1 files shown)


kpoeck:
Testing here:
```lisp
COMMON-LISP-USER> (integer-decode-float #.ext:short-float-positive-infinity)

Condition of type: ARITHMETIC-ERROR
Condition of type ARITHMETIC-ERROR was signaled.

Available restarts:
(use :r1 to invoke restart 1)

Broken at frame[0] (LAMBDA NIL). In: #<PROCESS TOP-LEVEL @0x10b0d25b9>.
COMMON-LISP-USER>> (core:btcl)
   0: ((LAMBDA NIL))
  25: (DEFAULT-DEBUGGER #S(ARITHMETIC-ERROR ))
  26: (INVOKE-DEBUGGER #S(ARITHMETIC-ERROR ))
  27: (UNIVERSAL-ERROR-HANDLER NIL ARITHMETIC-ERROR NIL)
  30: (ERROR ARITHMETIC-ERROR)
  40: (SIMPLE-EVAL (INTEGER-DECODE-FLOAT inf) NIL #S(STANDARD-GENERIC-FUNCTION CCLASP-EVAL-WITH-ENV #slots[11]))
  41: (CCLASP-EVAL (INTEGER-DECODE-FLOAT inf) NIL)
  42: (CATCH-LAMBDA)
  44: (TPL)
  45: (CATCH-LAMBDA)
````

so far so good, but:
```lisp
COMMON-LISP-USER>> :r1
Debugger received error of type: OUT-OF-EXTENT-UNWIND
Attempted to return or go to an expired block or tagbody tag.
Error flushed.
COMMON-LISP-USER>> (core:quit)
../../src/main/main.cc:264 There was an unhandled std::exception in process [pid: 79374] e.what()=[std::exception] - do something about it.
Abort trap: 6
````

I believe this is a additional error and has nothing to do with my change
@Bike can you verify this?

Bike:
Filed a bug: #961

PR #960, opened 2020-03-30
Define ./waf test and ./waf tests to execute the regression tests 

kpoeck:
* defines `./waf test` and `./waf tests` to execute the regression tests with current clasp
* assures that build/clasp is created, even if faso files are not used


Bike:
so just to be clear - if an executable _is_ created, `build/clasp` will still be it or point to it, right?

kpoeck:
Sorry not being more clear. With my change build/clasp is created, even for the non-faso case. Before the change build/clasp was only created in the faso case.

PR #964, opened 2020-04-02
Feature complete eclector character names

kpoeck:
* all char-names can be read in again (need to teach eclector them)
* reg test all-char-names-can-be-read-again for this
* regtest for issue-780 (fixed by bike in CST)
* fixes ansi tests `SYNTAX.SHARP-BACKSLASH.5, SYNTAX.SHARP-BACKSLASH.6`


Bike:
Do you think we should just bite the bullet and parse character names out of UnicodeData.txt?

kpoeck:
Looking at https://www.unicode.org/Public/UCD/latest/ucd/UnicodeData.txt we probably should.
Looking at the file I see the following:
* Should be very easy to parse
* First column has the character-code in HEX
* Second column has the Name, unless it is something like <char+>
* Codes 0000-001F have just <control>, but in clasp we have names
* Codes 007F-009F has just <control>, in clasp 007F is "Rubout",
* Codes 00A0-00FF have names in Unicode, but not in clasp
* 34 further lines have <something>
* In the names in the unicode file we need to substitute #\Space with #\_
* We need to cover cl:char-name, cl:name-char both in c++ and eclector.reader:find-character

Questions
* Should we start using the unicode table starting with 00A0 or even 0100?
* Do we just do this in lisp or do we teach c++ the same table?
* Do we also want to use the unicode data in the following 3 function
  * upper-case-p
  * lower-case-p
  * both-case-p

kpoeck:
Did now properly integrate the unicode Data file.
* process-unicode-file lazily loads the original data file rc/lisp/kernel/cleavir/UnicodeData.txt at run time, when it is first needed. Building clasp is therefor not impacted
* all defined unicode name starting from code #XA0 are loaded (about 34.000 entries). Updating should be as easy as substituting the file, no process needed
* names used in cl:name-char, cl:char-name, and eclector (in eclector.reader:find-character)
* write_character also uses this data calling out to lisp so that the unicode names are printed, e.g. `(code-char 300) -> #\LATIN_CAPITAL_LETTER_I_WITH_BREVE`. `#\LATIN_CAPITAL_LETTER_I_WITH_BREVE` can also be directly used
* Did run the regression-tests, added new tests that I found failing
* Did run the ansi-tests, until no regressions where found.
* As agreed, I did no mess with the *case-p*  functions

Please check

kpoeck:
@Bike , - when you have time - could you please check again whether this fits?

Bike:
I think we should keep the Unicode data in tools-for-build or something, not the cleavir directory.

And I don't know about doing it lazily... I'd prefer a longer build time to `char-name` taking a lot of time once. Actually, won't it be a longer build time anyway since we use the Eclector reader during build?

I think it would be good if during build we parse the unicode data into a format we can load quickly at startup.

Also, `name-char` doing a linear search is going to suck, but I guess we can worry about that later

kpoeck:
I am a bit puzzled about the comment about name-char, but probably I am missing something:

Name-char calls `eclector.reader:find-character` which does 
* a gethash `(gethash name *character-names*)`
* if the previous thing fails does another gethash `(gethash name *additional-clasp-character-names*)`

There is no linear search here (there are a-lists, but they are transferred to hash-tables.

The unicode file I can certainly move to tools-for-build and I can clearly preprocess the unicode file, but the lazy load is quick right now (less than 1 second).

Anyhow I wonder where to put the preprocessed file

could I simply do a new file with
```lisp
(eval-when (:compile-toplevel)
       (process-unicode-file))
(setq *mapping-char-code-to-char-names* #.*mapping-char-code-to-char-names*)
(setq *additional-clasp-character-names* #.*additional-clasp-character-names*)
````

Please advise

Bike:
Oh, I thought `name-char` was using the alist directly. My bad. SBCL has this complicated tree search thing which is what I was thinking of, but I guess a big hash table is probably fine too.

As for the file. So `preprocess-unicode-file` defines those variables? And the values of them are hash tables? Seems fine then.

kpoeck:
I am now trying the #.variable trick I refered earlier. Seems to compile real quick (1 second). As soon as tests confirm that lazy loading is not needed I update the pr for - hopefully -final review

kpoeck:
Hopefully final changes:
* did move the unicode data file to tools-for-build/UnicodeData.txt
* no lazy initialisation at runtime, all objects are created at build time with the compiler (see define-unicode-tables.lisp)
* did run again (with success)
  * regression test
  * ansi-tests

PR #965, opened 2020-04-03
define stat,fast (size,mtime.mode)

kpoeck:
to use, e.g.
```lisp
(let* ((file "~/quicklisp/setup.lisp"))
   (core:stat file))

5054
1464206628
33188
(let* ((file "~/quicklisp/setup.lisp")
       (fd nil))
  (with-open-file (stream file :direction :input)
    (setq fd (core::file-stream-fd stream))
  (core:fstat fd)))

5054
1464206628
33188
````

### DIFF
--- a/src/core/unixfsys.cc
+++ b/src/core/unixfsys.cc
@@ -2221,5 +2221,79 @@ CL_DEFUN FdSet_sp core__make_fd_set() {
   return fdset;
 }
 
+CL_LAMBDA(filedescriptor);
+CL_DECLARE();
+CL_DOCSTRING("Fstat-size for a file-descriptor, nil on error");
+CL_DEFUN T_sp core__fstat_size(int filedescriptor) {
+  struct stat sb;
+  if (fstat(filedescriptor, &sb) == -1)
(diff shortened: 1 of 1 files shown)


Bike:
This assumes POSIX, no? I mean, we pretty much do that anyway, but maybe we should put POSIX bindings in their own package or something.

And couldn't we just have one `fstat` function that returns three values or a structure or something?

kpoeck:
so fstat function and stat function each returning 3 values? That I can do.

kpoeck:
Updated to define:
* core:stat (pathname) returning (values st_size st_mtime st_mode)
* core:fstat (file-descriptor) returning (values st_size st_mtime st_mode)

kpoeck:
* I need these to implement https://github.com/xach/zacl so that  https://github.com/gendl/aserve works for clasp
* some single-threaded tests already work

kpoeck:
* portable aserve (in the version mentionned above) works now, will publish

Bike:
We should probably export these... but we should think about how to present a coherent API first, so I'll merge now and we can worry about that later.

PR #966, opened 2020-04-05
Avoid doble quotes for pathnames

kpoeck:
* Before they would print like `Loading #P""/Users/karstenpoeck/lisp/compiler/clasp-karsten/build/boehm/fasl/aclasp-boehm-bitcode/src/lisp/kernel/lsp/defstruct.faso""`
* Reason is the change in __repr__ for strings


drmeister:
Thank you.

PR #974, opened 2020-04-10
Move unicode code as drmeister requested

kpoeck:
* Moved code (unchanged) from ` src/lisp/kernel/cleavir/translate.lisp ` to  `src/lisp/kernel/cleavir/define-unicode-tables.lisp`


kpoeck:
* did rerun the regression-tests

drmeister:
Thank you

PR #976, opened 2020-04-12
Fix remf where the keys are not symbols, e.g. numbers

kpoeck:
* `(let ((place (list 9000 23)))(remf place 9000))` complaining that 9000 is not a symbol (true, but not required)
* added regression test


drmeister:
Thank you!

PR #978, opened 2020-04-15
Feature enable disable signals

kpoeck:
Implements control of signal handling from lisp
* register a default-handler for sigpipe, so that clasp does not exit
* Implements `(ext:ignore-interrupt <signal>)` to ignore a signal, e.g. `(ext:ignore-interrupt :sigpipe)`
* Implements `(ext:default-interrupt <signal>)`to reset to the default signal handler, e.g. `(ext:default-interrupt :sigpipe)`
* Implements `(ext:enable-interrupt <signal> <mode> &optional <function>)`, e.g. `(ext:enable-interrupt :sigpipe :lisp #'karsten-handler)` to define a handler in lisp. Mode can be
  * :ignore as in `(ext:ignore-interrupt <signal>)`  
  * :default as in `(ext:default-interrupt <signal>)`
  * :lisp as in `(ext:enable-interrupt :sigpipe :lisp #'karsten-handler)`
* `(ext:get-signal-handler <signal>)`, e.g. `(ext:get-signal-handler :sigpipe)` to get a lisp-handler for signal, if defined
* `(ext:set-signal-handler <signal> <handler>)` as an alternative interface - like ecl - for `(ext:enable-interrupt <signal> :lisp <handler>`
* Defined signals:
  * :SIGHUP  1
  * :SIGINT  2 
  * :SIGQUIT  3 
  * :SIGILL  4 
  * :SIGTRAP  5
  * :SIGABRT  6 
  * :SIGPOLL  7 
  * :SIGIOT  6 
  * :SIGEMT  7 
  * :SIGFPE  8
  * :SIGKILL  9 
  * :SIGBUS  10
  * :SIGSEGV  11 
  * :SIGSYS  12
  * :SIGPIPE  13
  * :SIGALRM  14 
  * :SIGTERM  15 
  * :SIGURG  16 
  * :SIGSTOP  17 
  * :SIGTSTP  18
  * :SIGCONT  19 
  * :SIGCHLD  20 
  * :SIGTTIN  21 
  * :SIGTTOU  22 
  * :SIGIO  23
  * :SIGXCPU  24
  * :SIGXFSZ  25 
  * :SIGVTALRM  26 
  * :SIGPROF  27
  * :SIGWINCH  28
  * :SIGINFO  29 
  * :SIGUSR1  30 
  * :SIGUSR2  31

### DIFF
--- a/src/gctools/interrupt.cc
+++ b/src/gctools/interrupt.cc
@@ -53,7 +53,10 @@ core::T_sp safe_signal_handler(int sig) {
   return key;
 }
 
-core::T_mv gctools__signal_info(int sig) {
+CL_LAMBDA(signal);
+CL_DECLARE();
+CL_DOCSTRING("return Current handler for signal");
+CL_DEFUN core::T_mv gctools__signal_info(int sig) {
   return Values(safe_signal_name(sig),safe_signal_handler(sig));
 }
 
@@ -281,7 +284,7 @@ CL_DEFUN void core__disable_all_fpe_masks() {
 
 CL_LAMBDA(&key underflow overflow inexact invalid divide-by-zero denormalized-operand);
 CL_DECLARE();
-CL_DOCSTRING("zerop");
+CL_DOCSTRING("core::enable-fpe-masks");
 CL_DEFUN void core__enable_fpe_masks(core::T_sp underflow, core::T_sp overflow, core::T_sp inexact, core::T_sp invalid, core::T_sp divide_by_zero, core::T_sp denormalized_operand) {
   // See https://doc.rust-lang.org/stable/core/arch/x86_64/fn._mm_setcsr.html
   // mask all -> no fpe-exceptions
@@ -402,7 +405,7 @@ void initialize_signals(int clasp_signal) {
     INIT_SIGNALI(SIGBUS, (SA_NODEFER | SA_RESTART), handle_bus);
   }
   INIT_SIGNALI(SIGFPE, (SA_NODEFER | SA_RESTART), handle_fpe);
-  // HAndle all signals that would terminate clasp (and can be caught)
+  // Handle all signals that would terminate clasp (and can be caught)
   INIT_SIGNAL(SIGILL, (SA_NODEFER | SA_RESTART), handle_signal_now);
   INIT_SIGNAL(SIGPIPE, (SA_NODEFER | SA_RESTART), handle_signal_now);
   INIT_SIGNAL(SIGALRM, (SA_NODEFER | SA_RESTART), handle_signal_now);
@@ -435,7 +438,10 @@ void initialize_signals(int clasp_signal) {
     ADD_SIGNAL_SYMBOL(sig,sigsym,handler); \
   }
 
-void gctools__push_unix_signal_handler(int signal, core::Symbol_sp name, core::Function_sp handler) {
+CL_LAMBDA(signal symbol function);
+CL_DECLARE();
+CL_DOCSTRING("Set current handler for signal");
+CL_DEFUN void gctools__push_unix_signal_handler(int signal, core::Symbol_sp name, core::Function_sp handler) {
   WITH_READ_WRITE_LOCK(_lisp->_Roots._UnixSignalHandlersMutex);
   ADD_SIGNAL_SYMBOL(signal,name,handler);
 }
--- a/src/gctools/interrupt.cc
+++ b/src/gctools/interrupt.cc
@@ -1,4 +1,5 @@
 #include <signal.h>
+#include <csignal>
 #include <xmmintrin.h>
 #include <llvm/Support/ErrorHandling.h>
 #include <clasp/core/foundation.h>
@@ -217,6 +218,20 @@ CL_DEFUN void core__check_pending_interrupts() {
   handle_all_queued_interrupts();
 }
 
+SYMBOL_EXPORT_SC_(CorePkg,call_lisp_symbol_handler);
+void lisp_signal_handler(int sig) {
+  core::eval::funcall(core::_sym_call_lisp_symbol_handler, core::clasp_make_fixnum(sig));
+}
+
+CL_DEFUN void core__enable_disable_signals(int signal, int mod) {
+  if (mod == 0)
+    std::signal(signal, SIG_IGN);
+  else if (mod == 1)
+    std::signal(signal, SIG_DFL);
+  else 
+    std::signal(signal, lisp_signal_handler);
+}
+      
 // HANDLERS
 
 // This is both a signal handler and called by signal handlers.
@@ -382,6 +397,9 @@ void initialize_signals(int clasp_signal) {
   }
   INIT_SIGNALI(SIGFPE, (SA_NODEFER | SA_RESTART), handle_fpe);
   INIT_SIGNAL(SIGILL, (SA_NODEFER | SA_RESTART), handle_signal_now);
+  INIT_SIGNAL(SIGPIPE, (SA_NODEFER | SA_RESTART), handle_signal_now);
+  // Ignore sigpipe for now
+  // signal(SIGPIPE, SIG_IGN);
   // FIXME: Move?
   init_float_traps();
   // core__disable_fpe_masks();
--- a/src/gctools/interrupt.cc
+++ b/src/gctools/interrupt.cc
@@ -56,7 +56,7 @@ core::T_sp safe_signal_handler(int sig) {
 CL_LAMBDA(signal);
 CL_DECLARE();
 CL_DOCSTRING("return Current handler for signal");
-CL_DEFUN core::T_mv gctools__signal_info(int sig) {
+CL_DEFUN core::T_mv core__signal_info(int sig) {
   return Values(safe_signal_name(sig),safe_signal_handler(sig));
 }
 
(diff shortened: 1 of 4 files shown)


kpoeck:
sample session in clasp and a terminal executing `kill -s PIPE <clasp-pid>`
* Default handler for sigpipe
`kill -s PIPE 52865`->
```lisp
COMMON-LISP-USER> 
Condition of type: UNIX-SIGNAL-RECEIVED
Serious signal 13 caught.

Available restarts:

1. (CONTINUE) Ignore signal

(PEEK-CHAR NIL #<SYNONYM-STREAM +PROCESS-STANDARD-INPUT+> NIL :EOF)
````
* Ignore sigpipe
`(ext:ignore-interrupt :sigpipe)`and `kill -s PIPE 52865` -> Nothing visible, signal ignored
* Define a lisp handler
`(defun karsten-handler(signal)(format t "~&Processed Signal ~a~%" signal))` and
`(ext:enable-interrupt :sigpipe :lisp #'karsten-handler)` and `kill -s PIPE 52865` ->
`Processed Signal 13`
* Get the handler:
`(ext:get-signal-handler :sigpipe)`  -> `#<FUNCTION KARSTEN-HANDLER>``
* Set the default handler:
`(ext:default-interrupt :sigpipe)`and `kill -s PIPE 52865` -> Clasp process terminates

kpoeck:
@Bike , when ever you have time, please review

Bike:
Docstrings for enable-interrupt, default-interrupt, and ignore-interrupt are not good. enable-interrupt should be more specific (like, say lisp-handler must not be provided unless `:lisp` is, and then it should be a function of one argument). default-interrupt doesn't "set the default handler", it sets the handler _to_ the default handler. ignore-interrupt doesn't ignore signals, it sets a handler that ignores signals.

Having `default-interrupt` and `ignore-interrupt` at all seems redundant, but i guess it's okay.

`*signal-to-function*` seems redundant - isn't it the same as the existing `_lisp->_Roots._UnixSignalHandlers` as used in interrupt.cc?

Don't use `std::signal`. C++'s specification is too vague (as explained on e.g. the Linux manpage). Use the POSIX function `sigaction` instead, like we do in interrupt.cc. Make sure that the flags include at least SA_NODEFER and SA_RESTART - the latter has to do with system calls, but the former is important in that it allows the signal to be received again while still in the signal handler, which is very possible if the handler aborts to the REPL or something.

We might want to allow Lisp signal handlers to get at the additional information sigaction can provide - but we can leave that for later.

Don't leave in commented out code like your karsten-handler unless it serves some specific purpose explained in a comment (e.g., as an example).

The signals list in `external-to-int-signal` may be problematic. It includes non-POSIX signals like WINCH. ([Here is a list of the POSIX-defined signals.](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/signal.h.html)) Those should be conditionalized based on the operating system. Additionally, the actual numbers of the signals may vary between operating systems - I don't know if they actually do, but it seems like something we might want to work out via C preprocessor flags.

Rather than just PIPE, all signals should have a handler, no? Or at least all the ones that terminate the process by default. These are marked "T" or "A" in the POSIX table.

kpoeck:
* using feature-enable-disable-signals  now
* limited the signals to those defined in posix and verified values for darwin, linux (x86 and arm) and freebsd
* improved comments as indicated
* deleted commented test code
* Added handlers for all signals that might terminate clasp (apart from SIGKILL and SIGSTOP that terminate clasp by definition)
* The following are handled in clasp
  * SIGINT
  * SIGQUIT
  * SIGILL
  * SIGTRAP
  * SIGABRT
  * SIGFPE
  * SIGBUS
  * SIGSEGV
  * SIGSYS
  * SIGPIPE
  * SIGALRM
  * SIGTERM
  * SIGURG
  * SIGTSTP
  * SIGCONT
  * SIGCHLD
  * SIGTTIN
  * SIGTTOU
  * SIGXCPU
  * SIGXFSZ
  * SIGXFSZ
  * SIGVTALRM
  * SIGPROF
  * SIGUSR1
  * SIGUSR2

kpoeck:
SIGUSR1 and SIGUSR2 seem to be used by clasp internally, so I wonder whether they should be allowed to be redefined by the user.

regarding _UnixSignalHandlers, they are used in handle_signal_now, but not all signals call handle_signal_now. There is obviously some overlap, but it is not the same, e.g. I coudn't overwrite sigfpe using this.

@Bike, unsure what to do with _UnixSignalHandlers, please advise. I believe I have covered all other comments

kpoeck:
* gctools:signal-info and gctools:push-unix-signal-handler where meant to be accessable from lisp
```lisp
(dotimes (x 31)
  (pprint (multiple-value-list (gctools::signal-info (1+ x)))))

(:SIGHUP NIL)
(:SIGINT CORE:TERMINAL-INTERRUPT)
(:SIGQUIT NIL)
(:SIGILL EXT:ILLEGAL-INSTRUCTION)
(:SIGTRAP NIL)
(:SIGABRT NIL)
(:SIGEMT NIL)
(8 8)
(:SIGKILL NIL)
(10 10)
(11 11)
(:SIGSYS NIL)
(:SIGPIPE NIL)
(:SIGALRM NIL)
(:SIGTERM NIL)
(:SIGURG NIL)
(:SIGSTOP NIL)
(:SIGTSTP NIL)
(:SIGCONT NIL)
(20 20)
(:SIGTTIN NIL)
(:SIGTTOU NIL)
(:SIGIO NIL)
(:SIGXCPU NIL)
(:SIGXFSZ NIL)
(:SIGVTALRM NIL)
(:SIGPROF NIL)
(:SIGWINCH NIL)
(:SIGINFO EXT:INFORMATION-INTERRUPT)
(:SIGUSR1 NIL)
(:SIGUSR2 NIL)
NIL
````

kpoeck:
and `gctools:push-unix-signal-handler` seems to require a function as last argument, but later code requires a symbol, fixing that

kpoeck:
* fixed `gctools:push-unix-signal-handler`
```lisp
(defun karsten()
  (format t "~&Karsten Handler~%"))
(gctools:push-unix-signal-handler 1 :sighup 'karsten)
;;; kill -s SIGHUP 75131 from a terminal
-> Karsten Handler
````

ok now consistent and ready for review

Bike:
So do `push-unix-signal-handler` and `enable-interrupt` do the same thing?

Less nitpicking now... `enable-interrupt` should probably signal an error if it's given something that doesn't actually name a signal. For the BSD signals, they should be conditionalized on a BSD feature, since there are posix systems other than linux and BSD, even if we don't support them right now.

kpoeck:
Signaling an error if the signal cannot be recognized is easy, testing now.
The are no bsd-specific signals in here, just verified again against the posix spec.
The only deviation to the posix spec is SIGPOLL, which is not defined on darwin nor on bsd, so I added it now with `#+linux`
My comment about bsd  was unfortunate, I only verified the bsd-mapping, but did not add any bsd-specific signals.

kpoeck:
Ok, now:
```lisp
(EXT:ENABLE-INTERRUPT :SIGNONEXISTANT :IGNORE)
->
Condition of type: SIMPLE-ERROR
Signal :SIGNONEXISTANT not recognized, Handler cannot be set

Available restarts:

````

@Bike, please check

Bike:
I didn't mean the BSD specific signals, I meant the numbers, like https://github.com/kpoeck/clasp/blob/feature-enable-disable-signals/src/lisp/kernel/lsp/posix.lsp#L73. The `#-linux` ones are FreeBSD, right?

kpoeck:
#-linux is #+(or darwin freebsd) (clasp also defines bsd for darwin)

I could do e.g. like
```lisp
...
(:SIGBUS  #+(or darwin freebsd) 10 #+linux 7 #-(or darwin freebsd linux) (error "Need to defined code for signal")    :term "bus error ")
(:SIGSEGV #+(or darwin freebsd linux) 11 #-(or darwin freebsd linux)  (error "Need to defined code for signal")     :core "segmentation violation ")
...
````

Bike:
Yeah, something like that. Or, could we define this mapping in C++ and use the actual macros?

kpoeck:
Seems like a sound proposal to define this alist in c++, that would be as safe a groveling, will do

kpoeck:
* did create the alist in c++:
```lisp
COMMON-LISP-USER> (core:signal-code-alist)

((:SIGINFO . 29) (:SIGWINCH . 28) (:SIGIO . 23) (:SIGEMT . 7) (:SIGUSR2 . 31)
 (:SIGUSR1 . 30) (:SIGPROF . 27) (:SIGVTALRM . 26) (:SIGXFSZ . 25)
 (:SIGXCPU . 24) (:SIGTTOU . 22) (:SIGTTIN . 21) (:SIGCHLD . 20)
 (:SIGCONT . 19) (:SIGTSTP . 18) (:SIGSTOP . 17) (:SIGURG . 16) (:SIGTERM . 15)
 (:SIGALRM . 14) (:SIGPIPE . 13) (:SIGSYS . 12) (:SIGSEGV . 11) (:SIGBUS . 10)
 (:SIGKILL . 9) (:SIGFPE . 8) (:SIGABRT . 6) (:SIGTRAP . 5) (:SIGILL . 4)
 (:SIGQUIT . 3) (:SIGINT . 2) (:SIGHUP . 1))
````
Since the following 2 are not really gc-related
* changed package for `gctools:push-unix-signal-handler` to `core:push-unix-signal-handler`and
* `gctools:signal-info`to `core:signal-info`

@Bike, hope thats all

Bike:
Great. Do you have permissions to edit the wiki? If so, please edit the manual with descriptions of these functions for when I forget about them in like a week

kpoeck:
yes have permissions, will do

PR #981, opened 2020-04-23
Expose ext:stat ext:fstat & ext:stream-file-descriptor + tests

kpoeck:
* (ext:stat <pathname>  &optional type) 
* (ext:fstat <file-descriptor>  &optional type) 
* (ext:stream-file-descriptor <stream>) 
* test for all above

Will add to manual, if ok like that

### DIFF
--- a/src/core/lispStream.cc
+++ b/src/core/lispStream.cc
@@ -4389,25 +4389,24 @@ BEGIN:
 
 CL_LAMBDA(s);
 CL_DECLARE();
-CL_DOCSTRING("core-file-stream-fd");
-CL_DEFUN T_sp core__file_stream_fd(T_sp s) {
-  T_sp ret;
+CL_DOCSTRING("Returns the file descriptor for a stream");
+CL_DEFUN T_sp ext__file_stream_file_descriptor(T_sp s) {
   unlikely_if(!AnsiStreamP(s))
-      FEerror("file_stream_fd: not a stream", 0);
-
+    TYPE_ERROR(s, cl::_sym_Stream_O);
+  T_sp ret;
   switch (StreamMode(s)) {
   case clasp_smm_input:
   case clasp_smm_output:
   case clasp_smm_io:
-    ret = make_fixnum(fileno(IOStreamStreamFile(s)));
-    break;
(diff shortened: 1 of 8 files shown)


kpoeck:
Forgot to add that I wanted these to implement clasp for https://github.com/ruricolist/trivial-file-size

Bike:
Why not just export `stat` from `ext:` directly?

kpoeck:
I assume you mean instead of defining core__stat directly define ext__stat?
Than a client has to do e.g. `(nth-value 0 (ext:stat namestring)` instead of `(ext:stat namestring :size)`

I like the second idiom a bit better, but not enough to discuss a lot. Your opinion?

If you look at 
https://github.com/ruricolist/trivial-file-size/blob/master/trivial-file-size.lisp#L35-#L51
gcl and cmucl do the nth -value thing, the others have a selector, probably since their stat returns a structure and they need the accessor

`core__file_stream_fd` is not used in any place in clasp, perhaps I could rename that to `ext__file_stream_fd`

Bike:
> Than a client has to do e.g. (nth-value 0 (ext:stat namestring)) instead of (ext:stat namestring :size)

I think that's fine, yeah. The selector thing strikes me as convoluted. For one thing they might want to use more than one property at a time, and having to repeatedly call the underlying `stat` function for that would be rather unfortunate.

As for `file-stream-fd`, seems fine conceptually. I'd probably verbosely name it `file-stream-file-descriptor` but whatever. And it could maybe use some cleanup - it signals a simple error if it doesn't get a file stream (instead of, like, a type error) and also does a printf for some reason.

kpoeck:
@Bike I believe I implemented all your comments, even renamed core__file_stream_fd to ext__file_stream_file_descriptor and changed the 2 places that did call core__file_stream_fd.

Did also cleanup `ext__file_stream_file_descriptor` as indicated:
```lisp
COMMON-LISP-USER> (ext:file-stream-file-descriptor 23)

Condition of type: TYPE-ERROR
23 is not of type STREAM.
````

kpoeck:
And added this test-case into the regression-test, so now really ready again

PR #985, opened 2020-04-26
Correctly define datatype of argument as number for cl__zerop

kpoeck:
* cl__zerop is defined to work on numbers, not need to pass a T_sp num and later do gc::As<Number_sp>(num). Don't even want to know what happens currently if you do `(zerop "23")`

* Did verify change with regression and ansi tests.


Bike:
`As` signals a type error, so there wouldn't actually be a problem. Specifying the type in the lambda list like this just does the same thing in the wrapper. Which is cleaner, so.

kpoeck:
thanks for the explanation

PR #986, opened 2020-04-29
Fix compile file serial faso

kpoeck:
Fix errors with the combination of 
CLASP_BUILD_MODE = 'faso'
USE_COMPILE_FILE_PARALLEL=False.

* In compile-file-output-module, respect the output-file
* in compile-file-serial use the same default :fasl for `output-type`as in compile-file-parallel
* added a regression-test

Validated by:
* running regression tests
* x-compiling sbcl (where the error was first noded)
* compiling the following quicklisp systems:
```lisp
(progn
  (time (ql:quickload :netcdf :verbose t))
  (time (ql:quickload :static-vectors :verbose t))
  (time (ql:quickload :trivial-garbage :verbose t))
  (time (ql:quickload :bordeaux-threads :verbose t))
  (time (ql:quickload :cffi :verbose t))
  (time (ql:quickload :usocket :verbose t))
  (time (ql:quickload :uuid :verbose t))
  (time (ql:quickload :trivial-backtrace :verbose t))
  (time (ql:quickload :esrap :verbose t)))
````
* quickloading mcclim

### DIFF
--- a/src/lisp/kernel/cmp/compile-file.lsp
+++ b/src/lisp/kernel/cmp/compile-file.lsp
@@ -88,12 +88,17 @@
     ((eq output-type :bitcode) (if *use-human-readable-bitcode* "ll" "bc"))
     ((and *generate-faso* (or (eq output-type :object))) "faso")
     ((and *generate-faso* (or (eq output-type :fasl))) "fasp")
+    #+no ((and (eq core:*clasp-build-mode* :faso) (or (eq output-type :object)))"faso")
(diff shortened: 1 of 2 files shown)


drmeister:
I don't know if this comment will modify things...

 *clasp-build-mode* is to control how the clasp builder builds clasp - I think it's somewhat unrelated to *generate-faso*.
*clasp-build-mode* tells aclasp/bclasp/cclasp what kind of object files they generate - either traditional .o (object) or .faso files.
*generate-faso* is what COMPILE-FILE will generate - and that is tied to whether *compile-file-parallel* is T or NIL.

PR #987, opened 2020-05-01
Fix FileStream_O::__repr__() for pipes

kpoeck:
* `(ext:vfork-execvp (list "llvm-config" "--ldflags" "--libdir" "--libs") t)` used to fail since `FileStream_O::__repr__()` tried to print the file position, but that fails for pipes
* now testing whether the file-position call can work before calling
* added a regression-test


kpoeck:
ready for review

Issue #961, opened 2020-03-30
Can't restart from floating point errors


Bike:
For example,
```
(handler-bind ((arithmetic-error #'abort))
  (restart-case
      (integer-decode-float #.ext:short-float-positive-infinity)
    (abort ())))
```
will signal an out of extent unwind. Something weird with signal handling, probably.

Bike:
The backtrace as obtained by `btcl` or `c_btcl` has the frame for the function that establishes the restart, but it still fails. Don't know how that's possible? The stack being copied? I really have no idea

Bike:
Seems to be just crashing now.

kpoeck:
with the change from longjmp to _longjmp I get again: (did also a fprintf in handle_fpe to see if the hander is called
```lisp
 (handler-bind ((arithmetic-error #'abort))
  (restart-case
      (integer-decode-float #.ext:short-float-positive-infinity)
    (abort ())))
Enter handle_fpe Signo: 8 Errno:0 Code:0

Condition of type: OUT-OF-EXTENT-UNWIND
Attempted to return or go to an expired block or tagbody tag.
Available restarts:
(use :r1 to invoke restart 1, etc.)

1. (ABORT) ABORT
2. (RESTART-TOPLEVEL) Go back to Top-Level REPL. ````

**Comment by Bike** [comment]
To elaborate: On linux I do

```
(defun foo (x y)
  (declare (single-float x y))
  (ext:with-float-traps-masked () (/ x y)))
(foo 1.0 0.0)
```

and it crashes while trying to print a backtrace for the division-by-zero.

Note that you can't just do `(/ 1.0 0.0)`, because LLVM will constant fold that into a NaN. (Which is something else we might need to rethink.)

kpoeck:
If I try this in main on macos I get:


```lisp
COMMON-LISP-USER> (defun foo (x y)
  (declare (single-float x y))
  (ext:with-float-traps-masked () (/ x y)))
; caught ERROR:
;   BUG: Bad rtype (SINGLE-FLOAT)
;     at unknown location
; 

Condition of type: SIMPLE-ERROR
BUG: Bad rtype (SINGLE-FLOAT)
Available restarts:
(use :r1 to invoke restart 1, etc.)

1. (RESTART-TOPLEVEL) Go back to Top-Level REPL.


(CLASP-CLEAVIR::INITIALIZE-IBLOCK-TRANSLATION #<IBLOCK NIL>)
[...]

PR #997, opened 2020-05-17
Define log1 for bignum, better Ratio_O::as_double_

kpoeck:
* replaces https://github.com/clasp-developers/clasp/pull/827
* defines
  * Bignum_O::log1_
  * Ratio_O::as_double_()
* allows:
```lisp
(labels ((factorial (n)(if (<= n 1) 1 (* n (factorial (1- n))))))
        (log (factorial 200)))

-> 863.232
````
* in cl-bench fixes:
  * pi-decimal/big
  * pi-atan
* fixes reading of very long float-tokens with eclector
  * see test very-long-floats-maxima and
  * test big-ratio-coerce
* read01.lisp has some additional tests not related to this change, that I has accumulated
* numbers.lisp has some disabled test regarding arithmetic errors that currently don't work because of https://github.com/clasp-developers/clasp/issues/961. Added them so they don't get lost
* maxima only works with this pr (and latest sicl changes from karlosz)

### DIFF
--- a/include/clasp/core/bignum.h
+++ b/include/clasp/core/bignum.h
@@ -245,8 +245,6 @@ public: // Functions here
   virtual bool evenp_() const { return (mpz_get_ui(this->_value.get_mpz_t()) & 1) == 0; };
   virtual bool oddp_() const { return (mpz_get_ui(this->_value.get_mpz_t()) & 1) != 0; };
 
-  Number_sp log1() const;
-
 }; // Bignum class
 
 }; // core namespace
--- a/include/clasp/core/bignum.h
+++ b/include/clasp/core/bignum.h
@@ -152,6 +152,7 @@ public: // Functions here
   Bignum get() const;
   Bignum get_or_if_nil_default(Bignum default_value) const;
   Number_sp abs_() const;
+  Number_sp log1_() const;
   Number_sp sqrt_() const;
   Number_sp reciprocal_() const;
   Number_sp rational_() const final { return this->asSmartPtr(); };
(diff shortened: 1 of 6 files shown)


Bike:
For the log1, the ratio constructed seems like it could have a 1 denominator sometimes. It's immediately converted to a float, but is this okay?

And for the mantissa thing, let me make sure I understand this correctly. When we want to convert a ratio to a float, we scale (multiply by `2^n`) the numerator such that the quotient is a whole number with as many digits as the mantissa of the float format, and then round it and use that scale factor in the exponent. Do I have that right? If so, since we're converting to a double float, shouldn't it be `DBL_MANT_DIG` instead of `FLT_MANT_DIG`? And when converting to a single float we just get a double and let C++ convert it to a single, so no problem there.

kpoeck:
The denominator is clasp_ash(clasp_make_fixnum(1), l). l is integer-length of a bignum. so at least 62. (ash 1 62) is already a bignum.

Nevertheless Rational_sp seems to be the better type for r (and if the denominator should be 1, clasp_make_rational takes care of that and returns a integer.

Additionally I seem to have missed a definition of Bignum_O::log1_ in numbers.cc, so dropping that.

I wonder if it is ok that Bignum_O::<definitions> are in numbers.cc?

Looking at `DBL_MANT_DIG`vs `FLT_MANT_DIG`now

Looking at

kpoeck:
There is Bignum_O::log1 and Bignum_O::log1_, confusing

kpoeck:
DBL_MANT_DIG and FLT_MANT_DIG are both 53 on macosx. But your argument is correct, so i change that to DBL_MANT_DIG

kpoeck:
There is also clasp_to_double( core::General_sp ); and clasp_to_double( core::T_sp ); but I tackle that separately

kpoeck:
@Bike: Now cleaned up and addressed your review comments, so please check again when you have time

PR #1000, opened 2020-05-19
Encodings: Drop big file with encoding data, add cache for requests

kpoeck:
* changed the generator to not create a big file with lisp expressions. Used technique in use for unicode-data. Compilation is now much, much faster 
* access to encodings is now cached
* did run regression-tests (no new regressions)


kpoeck:
@bike whenever you have time

PR #1001, opened 2020-05-21
Update eclector integration

kpoeck:
* create clasp-cleavir::clasp-eclector-client-mixin), so that for cst and non-cst functionality common behaviour can be defined once
* assure that (copy-readtable nil) still has the eclector readtable definitions
* update eclector via wscript 
* avoid &optional in eclector for return-float, declare inline (temporary until karlosz improves clearvir to handle this case)
* did run regression tests
* did run ansi-tests


kpoeck:
Also fixes: cl-test::print.complex.random.2 from ansi-tests (gives a abort trap 6) w/o this pr

PR #1004, opened 2020-05-22
[Wip] Drop unneeded declarations for numbers

kpoeck:
* drop most variations of clasp_to_double, had repetitive code
* unfortunately `double clasp_to_double( core::T_sp );` is used from cando, so coudn't drop that and clasp_to_double neds ctasts in some places
* `gc::Fixnum get_fixnum(Fixnum_sp x);` is not used, so deleted
* `clasp_deliver_fpe` is not used, so deleted
* ` LongFloat           clasp_to_long_float`and `LongFloat           clasp_to_long_double` where declared twice, so dropped the second

* did compile with cando, so assure that I don't delete anything used there

### DIFF
--- a/include/clasp/core/array_double.h
+++ b/include/clasp/core/array_double.h
@@ -24,8 +24,8 @@ namespace core {
     typedef specialized_SimpleVector_double TemplatedBase;
   public:
     static value_type default_initial_element(void) {return 0.0;}
-    static value_type from_object(T_sp obj) { return clasp_to_double(obj);};
-    static T_sp to_object(const value_type& v) { return core::clasp_make_double_float(v); };
+    static value_type from_object(T_sp obj) { return clasp_to_double(gc::As<Number_sp>(obj));};
+    static T_sp to_object(const value_type& v) { return clasp_make_double_float(v); };
   public:
   SimpleVector_double_O(size_t length, value_type initialElement=value_type(),
                           bool initialElementSupplied=false,
--- a/include/clasp/core/array_double.h
+++ b/include/clasp/core/array_double.h
@@ -24,7 +24,7 @@ namespace core {
     typedef specialized_SimpleVector_double TemplatedBase;
   public:
     static value_type default_initial_element(void) {return 0.0;}
-    static value_type from_object(T_sp obj) { return clasp_to_double(gc::As<Number_sp>(obj));};
+    static value_type from_object(T_sp obj) { return clasp_to_double(gc::As_unsafe<Number_sp>(obj));};
     static T_sp to_object(const value_type& v) { return clasp_make_double_float(v); };
   public:
   SimpleVector_double_O(size_t length, value_type initialElement=value_type(),
--- a/src/core/hashTable.cc
+++ b/src/core/hashTable.cc
@@ -235,7 +235,7 @@ CL_DEFUN T_sp cl__make_hash_table(T_sp test, Fixnum_sp size, Number_sp rehash_si
     }
     SIMPLE_ERROR(BF("Only :weakness :key (weak-key hash tables) are currently supported"));
   }
-  double rehash_threshold = maybeFixRehashThreshold(clasp_to_double(orehash_threshold));
+  double rehash_threshold = maybeFixRehashThreshold(clasp_to_double(gc::As<Number_sp>(orehash_threshold)));
   HashTable_sp table = _Nil<HashTable_O>();
   size_t isize = clasp_to_int(size);
   if (isize==0) isize = 16;
--- a/src/core/num_co.cc
+++ b/src/core/num_co.cc
@@ -113,10 +113,10 @@ CL_DEFUN Float_sp cl__float(Real_sp x, T_sp y) {
   case number_Ratio:
     switch (ty) {
     case number_SingleFloat:
-      x = gc::As<Real_sp>(clasp_make_single_float(clasp_to_double(x)));
+        x = gc::As<Real_sp>(clasp_make_single_float(clasp_to_double(gc::As<Number_sp>(x))));
       break;
     case number_DoubleFloat:
-      x = gc::As<Real_sp>(clasp_make_double_float(clasp_to_double(x)));
+        x = gc::As<Real_sp>(clasp_make_double_float(clasp_to_double(gc::As<Number_sp>(x))));
       break;
 #ifdef CLASP_LONG_FLOAT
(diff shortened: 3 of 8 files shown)


include/clasp/core/array_double.h
```diff
@@ -24,8 +24,8 @@ namespace core {
     typedef specialized_SimpleVector_double TemplatedBase;
   public:
     static value_type default_initial_element(void) {return 0.0;}
-    static value_type from_object(T_sp obj) { return clasp_to_double(obj);};
-    static T_sp to_object(const value_type& v) { return core::clasp_make_double_float(v); };
+    static value_type from_object(T_sp obj) { return clasp_to_double(gc::As<Number_sp>(obj));};
```
drmeister, line 27: This should use gc::As_unsafe<Number_sp>(obj). This would install a type check whenever from_object is called and we don't want them automatically installed everywhere.

src/core/hashTable.cc
```diff
@@ -235,7 +235,7 @@ CL_DEFUN T_sp cl__make_hash_table(T_sp test, Fixnum_sp size, Number_sp rehash_si
     }
     SIMPLE_ERROR(BF("Only :weakness :key (weak-key hash tables) are currently supported"));
   }
-  double rehash_threshold = maybeFixRehashThreshold(clasp_to_double(orehash_threshold));
+  double rehash_threshold = maybeFixRehashThreshold(clasp_to_double(gc::As<Number_sp>(orehash_threshold)));
```
drmeister, line 238: Here it's ok to use gc::As<Number_sp> because it's not time critical code and we want the type to be checked.

src/core/num_co.cc
```diff
@@ -113,10 +113,10 @@ CL_DEFUN Float_sp cl__float(Real_sp x, T_sp y) {
   case number_Ratio:
     switch (ty) {
     case number_SingleFloat:
-      x = gc::As<Real_sp>(clasp_make_single_float(clasp_to_double(x)));
+        x = gc::As<Real_sp>(clasp_make_single_float(clasp_to_double(gc::As<Number_sp>(x))));
```
drmeister, line 116: We know x is already a Number_sp because it's passed as a Real_sp
So use gc::As_unsafe<Number_sp>(x)

src/core/num_co.cc
```diff
@@ -113,10 +113,10 @@ CL_DEFUN Float_sp cl__float(Real_sp x, T_sp y) {
...
       break;
     case number_DoubleFloat:
-      x = gc::As<Real_sp>(clasp_make_double_float(clasp_to_double(x)));
+        x = gc::As<Real_sp>(clasp_make_double_float(clasp_to_double(gc::As<Number_sp>(x))));
```
drmeister, line 119: We know x is already a Number_sp because it's passed as a Real_sp
So use gc::As_unsafe<Number_sp>(x)

src/core/num_co.cc
```diff
@@ -399,7 +399,7 @@ Real_mv clasp_floor2(Real_sp x, Real_sp y) {
     }
     break;
   case number_SingleFloat: { /* SF / ANY */
-    float n = clasp_to_double(y);
+    float n = clasp_to_double(gc::As<Number_sp>(y));
```
drmeister, line 402: Here y is a T_sp - so yes - use gc::As<Number_sp>(y)

src/core/numbers.cc
```diff
@@ -694,21 +694,21 @@ CL_DEFUN Number_sp contagen_div(Number_sp na, Number_sp nb) {
   case_Bignum_v_Complex : {
       Complex_sp cb = gc::As<Complex_sp>(nb);
       return complex_divide(clasp_to_double(na), 0.0,
-                            clasp_to_double(cb->real()), clasp_to_double(cb->imaginary()));
+                            clasp_to_double(gc::As<Number_sp>(cb->real())), clasp_to_double(gc::As<Number_sp>(cb->imaginary())));
```
drmeister, line 697: cb->real() and cb->imaginary() return Real_sp types.  These inherit from Number_sp - so don't use gc::As<Number_sp>(cb->real()).   

clasp_to_double(cb->real()) should be fine.
If you want to be explicit use clasp_to_double(gc::As_unsafe<Number_sp>(cb->real()))

src/core/primitives.cc
```diff
@@ -115,7 +115,7 @@ CL_DOCSTRING("sleep");
 CL_DEFUN T_sp cl__sleep(Real_sp oseconds) {
   // seconds - a non-negative real.
   SYMBOL_EXPORT_SC_(ClPkg, sleep);
-  double dsec = clasp_to_double(oseconds);
+  double dsec = clasp_to_double(gc::As<Number_sp>(oseconds));
```
drmeister, line 118: As above oseconds is a Real_sp - that inherits from Number_sp so...
clasp_to_double(oseconds) is correct

src/core/numbers.cc
kpoeck: clasp_to_double(cb->real()) is not fine, the compiler is telling that type is ambigious and it doesn't know whether to call clasp_to_double for number or for T, so I will gc::As_unsafe<Number_sp> instead

src/core/num_co.cc
kpoeck: done

src/core/num_co.cc
kpoeck: ok

src/core/primitives.cc
kpoeck: compiler says:
../../src/core/primitives.cc:118:17: error: call to 'clasp_to_double' is ambiguous
  double dsec = clasp_to_double(oseconds);
                ^~~~~~~~~~~~~~~
../../include/clasp/core/numbers.fwd.h:51:11: note: candidate function
   double clasp_to_double( core::T_sp );
          ^
../../include/clasp/core/numbers.h:1090:23: note: candidate function
  double              clasp_to_double( core::Number_sp );
                      ^
1 error generated.

will use  double dsec = clasp_to_double(gc::As_unsafe<Number_sp>(oseconds));

kpoeck:
@drmeister , should have done all what you requested, please check

kpoeck:
ccando-boehm also starts fine

kpoeck:
@drmeister could you please recheck?

kpoeck:
substituted by #1060

PR #1010, opened 2020-05-27
Update lambda list parsing to match changes in CST.

karlosz:
Depends on https://github.com/s-expressionists/Concrete-Syntax-Tree/pull/27.


Bike:
Have to update our CST commit as well, presumably.

karlosz:
Updated.

PR #1012, opened 2020-05-28
Fix generate encodings not use read from string

kpoeck:
* replace potentially expensive read-from-string (signals in eclector) with cheaper operation
* compiles much quicker in my tests


drmeister:
Thank you

drmeister:
Whoops - commenting and then closing before merging is the wrong way to do this.

Issue #1014, opened 2020-05-30
(fboundp '(setf nil)) -> Type-error


kpoeck:
```lisp
(fboundp '(setf nil)) -> 
(SETF NIL) is not of type (SATISFIES CORE:VALID-FUNCTION-NAME-P).
   [Condition of type TYPE-ERROR]
````

kpoeck:
my own mistake, testing a fix.
Happens during use of :clasp-doc

Bike:
has been bothering me for a while but not enough to untangle function-block-name

PR #1019, opened 2020-05-31
Fix (fboundp '(setf nil)) signalling an error

kpoeck:
* (fboundp '(setf nil)) incorrectly returned an error, now simply nil
* did dropped some useless parenthesis in the c++ code
* extended functionBlockName so that we can distinguish a BlockName of nil from and error indicated by nil and used that in the 2 callers
* added regression-tests
* did run regression-tests and ansi-tests


kpoeck:
fixes #1014

PR #1022, opened 2020-06-07
[Wip] Allow to regenerate the file lists even from an installed clasp

### LINKED ISSUE #1020: regenerate-cleavir-file-list bases paths off the current executables source tree
It would be better to grovel based on the pathname of the file being loaded, otherwise using an executable installed into a different location will not use the intended paths.

kpoeck:
* fixes #1020 
* allows to regenerate the lists even from an installed clasp

### DIFF
--- a/tools-for-build/regenerate-cleavir-file-list.lisp
+++ b/tools-for-build/regenerate-cleavir-file-list.lisp
@@ -26,6 +26,8 @@
 
 (defun save-file-list-as-python (filename file-list)
   "Save the list of files in FILE-LIST to FILENAME as a python file."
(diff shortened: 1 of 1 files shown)


karlosz:
This also needs to be made independent of the logical translation for SYS:.

kpoeck:
@karlosz 
are you refering to this?
```lisp
(asdf:initialize-source-registry
 `(:source-registry
   (:tree ,(translate-logical-pathname #P"sys:"))
   :ignore-inherited-configuration))
````

kpoeck:
@karlosz please check again

PR #1023, opened 2020-06-07
Fix lisp bugs for weak-hash-tables

kpoeck:
* I understand that we might not want to advertise weak hash tables
* But I still would like to fix obvious errors
* replaces #947


kpoeck:
The changes in foundation.lisp are no longer needed, but the one in inline still

kpoeck:
Did drop the changes in foundation.lisp, so now mergeable

PR #1029, opened 2020-06-28
Correct wrong throw

kpoeck:
* See https://trac.clozure.com/ccl/ticket/1085 reported by stassats 7 years ago (wrong in ccl, ecl, clasp)
* WIll fix in ccl and ecl too


Bike:
I'm gonna just remove the code size estimation entirely.

kpoeck:
Now I have to make a new pr with the regression test, o well

PR #1048, opened 2020-09-06
Make make-array more robust + regression-tests

kpoeck:
* better (type)-error in calculateArrayTotalSizeAndValidateDimensions
* more type-checks in make-array
* regression-tests for all (and they found more errors) with and w/o the compiler-macro


kpoeck:
* hope I can use `(error 'type-error :datum ..` in make-array
* did run regression-tests with no new issues 
* did run ansi-test with no new issues

drmeister:
Thank you!

PR #1071, opened 2020-10-24
Added bounds checks for elt, have all errors use CORE:SEQUENCE-OUT-OF-BOUNDS

kpoeck:
* Cons_O::elt already had a bounds check, just but the correct error
* Cons_O::setf_elt dito
* for the vector case, put checks in
  * cl__elt
  * core__setf_elt
* fix error message for (elt nil <an-index>), nil is a sequence
* and the same for (setf (elt ...))


kpoeck:
Testing with ansi-test show that (elt nil index) gets the wrong error, adding this as well

drmeister:
Thank you!

PR #1079, opened 2020-11-02
Use Eclectors without-signal branch

kpoeck:
* as discussed on #sicl : use the without-signal branch (has been rebased and contains all latest fixes)
* tested with
  * regression tests
  * ansi-tests
  * icando-boehm

### DIFF
--- a/wscript
+++ b/wscript
@@ -295,8 +295,8 @@ def update_dependencies(cfg):
                        "https://github.com/robert-strandh/Acclimation.git",
                        "dd15c86b0866fc5d8b474be0da15c58a3c04c45c")
     fetch_git_revision("src/lisp/kernel/contrib/Eclector",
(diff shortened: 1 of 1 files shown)


Bike:
I think it would be better to target this to the master branch?

kpoeck:
I believe I am simply not aware of a good criteria to which branch to go.

To branch bir only things that actually have to do with BIR? Then this pr needs to go to master as you say.

Please advise

Bike:
My criterion is that there's a hopefully small chance BIR doesn't work and I chuck the whole thing, and in that eventuality I'd like things like this that would be useful regardless to remain. So yes, things that don't have to do with BIR should go to master. I can just rebase them into BIR.

kpoeck:
ok will do as you said, so closing this and creating a new pr to master

PR #1082, opened 2020-11-03
Add regression test for unwinding.

karlosz:
Make sure we don't unwind while compiling. (This is really a check on Eclector being optimized well.)

### DIFF
--- a/src/lisp/regression-tests/run-all.lisp
+++ b/src/lisp/regression-tests/run-all.lisp
@@ -39,6 +39,7 @@
 (load-if-compiled-correctly "sys:regression-tests;control01.lisp")
 (load-if-compiled-correctly "sys:regression-tests;loop.lisp")
 (load-if-compiled-correctly "sys:regression-tests;numbers-core.lisp")
+(load-if-compiled-correctly "sys:regression-tests;unwind.lisp")
 #+unicode
 (load-if-compiled-correctly "sys:regression-tests;encodings.lisp")
 (load-if-compiled-correctly "sys:regression-tests;system-construction.lisp")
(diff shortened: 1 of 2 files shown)


Bike:
this may be tangential, but doesn't this test dump a fasl in the source directory? might be good to put it elsewhere

kpoeck:
What about:
`(cmp::compile-file-serial "sys:kernel;lsp;predlib.lsp" :output-file (make-pathname :type (pathname-type (compile-file-pathname "foo.lisp")) :defaults (core:mkstemp "TMP:predlib")))`
instead of
`(cmp::compile-file-serial "sys:kernel;lsp;predlib.lsp")`

kpoeck:
and the same question whether this should go to bir or master

kpoeck:
But worse, test fails for me, I get 6 unwinds:
```lisp
(time (cmp::compile-file-serial
         "sys:kernel;lsp;predlib.lsp"
         :output-file (make-pathname
                       :type (pathname-type (compile-file-pathname "foo.lisp"))
                       :defaults (core:mkstemp "TMP:predlib"))))
....
Writing temporary bitcode file to: #P"/private/tmp/predlib3njLHN.ll"
Writing faso file to: #P"/private/tmp/predlib3njLHN.fasp"
Time real(10.066 secs) run(10.067 secs) consed(621569536 bytes) unwinds(6)
#P"/private/tmp/predlib3njLHN.fasp"
````

karlosz:
It works with the latest bir commit.

kpoeck:
Probably my error, so retesting

kpoeck:
works fine now, was an error on my part

Bike:
Oh yeah, we have `mkstemp`? works for me

karlosz:
Going to add mkstemp and lose this PR in favor of another one with more stuff in it.

PR #1083, opened 2020-11-06
Take into account that the standard requires the result of sqrt be a single float (see 12.1.3.3 Rule of Float Substitutability)

kpoeck:
as discussed on #clasp with Bike


Bike:
Later on this might signal an overflow instead of returning infinity. I suppose a good test might just be to check that the result, if it exists, is a single float?

drmeister:
Shall we merge this?

PR #1084, opened 2020-11-08
Fix handling of compiler conditions, so that source information is reported

kpoeck:
* The parallel compiler did not use the compiler condition infrastructure properly, so that  
  * compiler-error-handler
  * compiler-style-warning-handler
  * compiler-warning-handler
  * were not called, and consequently source positions were not reported (since print-compiler-condition was not used)

* fixed that `(cmp:compiler-condition-origin cleavir-conditions:program-condition)`crashed since it expected a cons of origins, but `clean-up-variable` just puts 1 origin
* To make explicit where we don't find origin information I added `    at unknown source` when we report a condition without origin

* what still is to be done, but i'd tackle that separately, is that `*compiler-error-count*` and friends don't communicate over threads, so the numbers reported in `compilation-unit-finished` are not correct

* `clean-up-variable` is just present in bir, thats why i developed all that in bir

### DIFF
--- a/src/lisp/kernel/cleavir/setup.lisp
+++ b/src/lisp/kernel/cleavir/setup.lisp
@@ -406,8 +406,12 @@ when this is t a lot of graphs will be generated.")
 (defmethod cmp:compiler-condition-origin
     ((condition cleavir-conditions:program-condition))
   ;; FIXME: ignore-errors is a bit paranoid
-  (ignore-errors (car (cleavir-conditions:origin condition))))
-
+  (let ((origin (cleavir-conditions:origin condition)))
+    (ignore-errors
+     (if (consp origin)
+         (car origin)
+         origin))))
+         
 (defun build-and-draw-ast (filename cst)
   (let ((ast (cleavir-cst-to-ast:cst-to-ast cst *clasp-env* *clasp-system*)))
     (cleavir-ast-graphviz:draw-ast ast filename)
--- a/src/lisp/kernel/cmp/compile-file-parallel.lsp
+++ b/src/lisp/kernel/cmp/compile-file-parallel.lsp
@@ -120,11 +120,12 @@
                     (progn
                       (cfp-log "Thread ~a compiling form~%" (mp:process-name mp:*current-process*))
                       (handler-case
-                          (funcall compile-func ast-job
-                                   :optimize optimize
-                                   :optimize-level optimize-level
-                                   :intermediate-output-type intermediate-output-type
-                                   :write-bitcode write-bitcode)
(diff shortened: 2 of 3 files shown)


Bike:
So it does `with-compilation-results` separately for each AST worker thread? Would it not be better if it was done around the overall process? Obviously this would take some thinking given the multiple kinds of conditions

kpoeck:
answering your question: yes for every worker thread. `call-with-compilation-results` does setup the handlers. I assumed i need todo this in every worker thread

Bike:
Let's review the behavior we want. If the compiler signals any kind of unhandled error or warning or style-warning, we want to increment the appropriate count, set `warnings-p` and/or `failure-p`, print the condition, and then suppress it. For compile file parallel, we need to do this both in the main thread, because it does work generating ASTs, and the worker threads, which do post-AST in BIR that until now never really signaled anything but bugs.

We do need signal handlers in each thread like you said, but what these handlers should do is a little more up in the air. The lack of source info is one problem, but we're also not handling `warnings-p` and `failure-p` correctly - if you c-f-p a file with an unused variable warning, compile-file incorrectly returns  that no warnings occurred (i.e. the secondary value is nil). I don't think your change here fixes that, since the values returned by each thread's `with-compilation-results` will be discarded.

We probably don't want each worker thread to print to the same error report stream, since then the output could get jumbled if multiple threads do so simultaneously.

Putting this all together, how does this sound: we give each AST job a slot for non-error conditions it catches (warnings, but probably also notes later). Once an AST job is complete, the main thread loops through the slot re-signaling each condition. This will be in the context of the main thread's `with-compilation-results` so things should work properly. We use `signal` and not `warn` because we want the `with-compilation-results` handlers doing the printing. This will also make `warnings-p` and `failure-p` sensible. I'll see about implementing this now.

Also, about AST job errors. What we do now is that if an AST job hits an error, it stores that error in the job and aborts, and then the main thread loops through the jobs signaling any saved errors. That might not actually be the best way to go. Errors like this can't reasonably be recovered from because the thread has already exited. But maybe I'll change that later.

Bike:
aaaah this cons versus not source info thing is so annoying though

Bike:
hm, actually notes will be tricky. the system i sketched out in cleavir relies on using its `note` function to get a report out. i'll just focus on warnings for now i guess

Bike:
https://github.com/clasp-developers/clasp/commit/06b4b503789007738c5b9c337c29df15b923ce5a there results in better reporting on my ened

kpoeck:
closing this, since bike included something similar, but more complete

PR #1087, opened 2020-11-14
Disable the float test that hang in BR, they work in master

kpoeck:
* for whatever reason they work in master and sometimes even in bir
* they hang in `contagen_div`
* specifically the follow patterns hangs `ext:with-float-traps-masked () ...)`

disable them until we find the reason


drmeister:
I'll merge this so we can get the tests to run through. 
I guess it's going to take more work to catch floating point errors.

PR #1089, opened 2020-11-15
Feature add native callback test

kpoeck:
* use clasps native ffi in the regression tests in order not to use quicklisp and cffi


drmeister:
Thank you very much!

PR #1090, opened 2020-11-16
Stack allocate closure stuff.

karlosz:


Bike:
I just merged this. Unfortunately, I got a little funky with rebases, so the commits have different hashes. The final one is d6c95f350f28cac2b2b1a03e078ad8e2f4f0cdb9.

PR #1092, opened 2020-11-17
Add some more standard library functions as DX-CALL.

karlosz:
Shaves about 15 mib of consing from compilation of asdf.


Bike:
Are all these changes fine? When the PR says one thing but the branch has more changes it's a bit ambiguous

karlosz:
Should be OK. Those were some other changes I pushed that didn't have anything to do with the PR but they all build fine with latest Cleavir.

PR #1105, opened 2020-12-12
updated llvm-config search path for 64 bit fedora 32

guicho271828:
according to https://github.com/mesonbuild/meson/issues/5059 and https://github.com/pexip/meson/commit/a4bd45fa1069dc87ec83640f9ddeb546d29ad8f5
it started on f30 (fedora 30), so they f30 and f31 will also be supported


drmeister:
Thank you!

Issue #1120, opened 2020-12-27
we don't handle char-downcase or char-upcase for unicode chars correctly


kpoeck:
as Bike notes in #164 

```lisp
(char-downcase #\CYRILLIC_CAPITAL_LETTER_IE_WITH_GRAVE)
-> #\CYRILLIC_CAPITAL_LETTER_IE_WITH_GRAVE
````

kpoeck:
In https://en.cppreference.com/w/cpp/locale/tolower and https://en.cppreference.com/w/cpp/locale/toupper

This works in Xcode, so lets try this for clasp

kpoeck:
seems to work fine:
```lisp
COMMON-LISP-USER> (char-downcase #\CYRILLIC_CAPITAL_LETTER_IE_WITH_GRAVE)
#\CYRILLIC_SMALL_LETTER_IE_WITH_GRAVE
COMMON-LISP-USER> (char-upcase #\CYRILLIC_SMALL_LETTER_IE_WITH_GRAVE)
#\CYRILLIC_CAPITAL_LETTER_IE_WITH_GRAVE
````

kpoeck:
Works now fine with pr #1121 merged

PR #1121, opened 2020-12-27
Implement char-downcase char-upcase correctly for unicode

kpoeck:
* implement proper char-downcase char-upcase for unicode
* regenerate…lower-case-p and upper-case-p from unicode data file
* tests


kpoeck:
solves #1120

Bike:
Does this construct a new locale object for each call? Does that use some kind of slow lookup?

kpoeck:
Possibly constructs a new locale on every call. Will:
* define the locale as a constant
* compare runtimes

kpoeck:
ok, now this is 1000 times quicker with the locale as a constant, @Bike please recheck

Bike:
I wonder if we might want to be able to customize the locale. But I guess we can worry about that later. UTF-8 is pretty much definitely better than just ASCII. I'll merge this after pushing some changes I have.

Issue #1141, opened 2021-02-15
Potential package error wrt import


kpoeck:
The following works fine in clasp, but errors in sbcl. I beleive that sbcl is correct:
```lisp
(in-package :cl-user)

(defpackage foo-bar-1
  (:use)
  (:export #:test))

(defpackage foo-bar-2
  (:use)
  (:export #:test))

(defpackage foo-bar-3
  (:use))

(import 'foo-bar-1:test (find-package :foo-bar-3))
(import 'foo-bar-2:test (find-package :foo-bar-3))

(do-symbols (sym (find-package :foo-bar-3))
  (print sym))
->
FOO-BAR-1:TEST 
NIL
````

kpoeck:
In sbcl this gives:
```lisp
...
* (import 'foo-bar-2:test (find-package :foo-bar-3))

debugger invoked on a NAME-CONFLICT in thread
#<THREAD "main thread" RUNNING {1004AB8143}>:
  IMPORT FOO-BAR-2:TEST causes name-conflicts in #<PACKAGE "FOO-BAR-3"> between
  the following symbols:
    FOO-BAR-2:TEST, FOO-BAR-1:TEST
See also:
  The ANSI Standard, Section 11.1.1.2.5

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [SHADOWING-IMPORT-IT] Shadowing-import FOO-BAR-2:TEST, uninterning FOO-BAR-1:TEST.
  1: [DONT-IMPORT-IT     ] Don't import FOO-BAR-2:TEST, keeping FOO-BAR-1:TEST.
  2: [RESOLVE-CONFLICT   ] Resolve conflict.
  3: [ABORT              ] Exit debugger, returning to top level.

(NAME-CONFLICT #<PACKAGE "FOO-BAR-3"> IMPORT FOO-BAR-2:TEST FOO-BAR-2:TEST FOO-BAR-1:TEST)
0] 
````

kpoeck:
I suspect clasp silently ignores the second import, see https://github.com/clasp-developers/clasp/blob/master/src/core/package.cc#L999

kpoeck:
Works fine in ECL too, t.m. a cerror is raised

kpoeck:
@Bike could you check, if this is really a bug, than I would work on this.

Bike:
[...]

PR #1152, opened 2021-03-25
Fix most package error from ansi-tests + new regression-tests

kpoeck:
* now only these 3 remains unsolved `feature-assorted-package-fixes
* solves #1141

### DIFF
--- a/include/clasp/core/designators.h
+++ b/include/clasp/core/designators.h
@@ -54,6 +54,8 @@ extern Path_sp pathDesignator(T_sp obj);
  
 /*! Return a Package by interpreting a package designator */
 extern Package_sp packageDesignator(T_sp obj);
+ /*! Return a Package by interpreting a package designator, or nil if not found */
+extern T_sp packageDesignatorNoError(T_sp obj);
 
 /*! Return the name of a Package by interpreting a package or a string as a name */
 extern string packageNameDesignator(T_sp obj);
--- a/src/core/designators.cc
+++ b/src/core/designators.cc
@@ -86,8 +86,8 @@ CL_DEFUN Function_sp coerce_fdesignator(T_sp obj) {
 
 namespace core {
 namespace coerce {
-core::Package_sp packageDesignator(core::T_sp obj) {
-  // TODO: Add support for Unicode package names
+core::T_sp packageDesignatorInternal(core::T_sp obj, bool errorp){
+    // TODO: Add support for Unicode package names
   String_sp packageName;
   if (Package_sp apkg = obj.asOrNull<Package_O>()) {
     return apkg;
@@ -111,7 +111,16 @@ core::Package_sp packageDesignator(core::T_sp obj) {
     Package_sp pkg = gc::As<Package_sp>(tpkg);
     return pkg;
   }
-  PACKAGE_ERROR(packageName);
+  if (errorp)
+    PACKAGE_ERROR(packageName);
+  else return _Nil<T_O>();
+}
+
+core::T_sp packageDesignatorNoError(core::T_sp obj) {
+  return packageDesignatorInternal(obj, false);
+}
+core::Package_sp packageDesignator(core::T_sp obj) {
+  return gc::As<Package_sp>(packageDesignatorInternal(obj, true));
 }
 };
 };
--- a/src/core/lisp.cc
+++ b/src/core/lisp.cc
@@ -975,19 +975,30 @@ Package_sp Lisp_O::makePackage(const string &name, list<string> const &nicknames
    * something, it will try to grab the lock, and it will fail because we already have it.
    * Additionally, CLHS specifies that the errors are CORRECTABLE, so we could hypothetically have the user
    * in the debugger while the package system is locked - that won't work.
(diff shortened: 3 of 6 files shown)


Bike:
By coincidence I have also been working on the package test failures, though I was focusing more on name conflicts than on package creation/deletion. I think some of your correctable errors here aren't totally correct - the CLHS is pretty specific about what restarts need to be able to do (11.1.1.2.5), but the ansi-tests can't cover that because the names of the restarts aren't standard. Many of those actions are things that would be pretty annoying to write in C++. For example, for `import`, "A name conflict in import with a symbol already present in the package may be resolved by uninterning that symbol, or by not doing the import."; since there are two possible actions this can't be done with just `cerror`.

Long story short, what I'm going to do is merge your changes into my working branch, then merge that branch. I'll discard some of your changes but keep the package designator stuff and the new tests.

Also, what's with this bit? https://github.com/clasp-developers/clasp/pull/1152/commits/3a6e0c85b9b6f9560f0c190efbaab521bc4cfdf1#diff-e83b41979710f43a5ef9eda0b8c6122b868de51904efa7423f2ebb06e571f6b7R902-R904 CLHS says in 11.1.1.2.3 that `import` does not change the home package of the importees.

kpoeck:
regarding the home package of uninterned symbols, clhs does say it.

http://www.lispworks.com/documentation/HyperSpec/Body/f_import.htm

"If any symbol to be imported has no home package (i.e., (symbol-package symbol) => nil), import sets the home package of the symbol to package."

Regarding the correctable errors I understand clhs differently, but Ithan  am no native speaker. I understand "may" as a possibility, not as a rule.

Bike:
ah, you're right about import. cool, the CLHS contradicts itself. setting the home package seems fine then.

Bike:
Does this work in your branch? It doesn't in my working branch that I've merged your changes into, and I'm not sure if it's something I did.

```lisp
(let ((par0 (make-package "PAR0")) (par1 (make-package "PAR1")) (par2 (make-package "PAR2")) (chil (make-package "CHIL")))
  (use-package (list par0 par1 par2) chil)
  (delete-package chil) (delete-package par0) (delete-package par1) (delete-package par2))
```

For me, the used-by lists are getting messed up somehow, so par0 can't be deleted because chil depends on it, even though chil has already been deleted.

kpoeck:
I get in my branch
```lisp
Condition of type: SIMPLE-PACKAGE-ERROR
Package with designator #<PACKAGE PAR0> is used by at least one other package:
Available restarts:

1. (CONTINUE) Remove dependency in other packages
````

kpoeck:
there might be an error even before my changes
```lisp
(defvar *par0*)
(defvar *par1*)
(defvar *par2*)
(defvar *chil*)

(setq *par0* (make-package "PAR0" :use nil)
      *par1* (make-package "PAR1" :use nil)
      *par2* (make-package "PAR2" :use nil)
      *chil* (make-package "CHIL11" :use nil))

(use-package (list *par0* *par1* *par2*) *chil*)
(delete-package *chil*)

(package-used-by-list *par0*) -> nil
(package-used-by-list *par1*) -> (#<PACKAGE >)
(package-used-by-list *par2*) -> nil
````
and unsurprisingly `(delete-package *par1*)` gives the cerror.

So my first suspicion is, the error was there before my changes, but now it is noted.

kpoeck:
I did put the following in Package_O::unusePackage_no_outer_lock:
```c++
printf("unuse Package %s from Package %s\n",
         usePackage->_Name->get_std_string().c_str(),
                 me->_Name->get_std_string().c_str()); 
````
this prints:for (delete-package *chil*)
unuse Package PAR0 from Package CHIL11
unuse Package PAR2 from Package CHIL11
unuse Package PAR2 from Package CHIL11

So in `Package_O::unusePackage_no_outer_lock` we do `this->_UsingPackages.erase(it);`
while in `cl__delete_package` we do 
```c++
 for (auto pi : pkg->_UsingPackages) {
    if (pi.notnilp())
      pkg->unusePackage_no_outer_lock(pi);
  }
`````

do we modify _UsingPackages while iterating over it, does not look like a great idea

kpoeck:
ok, changed the code in cl__delete_package to:
```c++
// need to grab before the readwriteLock
  List_sp usedList = pkg->packageUseList();
  List_sp usedByList = pkg->packageUsedByList();
  
  WITH_PACKAGE_READ_WRITE_LOCK(pkg);
  for (auto pi : usedList) {
    Package_sp piLocal = gc::As<Package_sp>(CONS_CAR(pi));
    pkg->unusePackage_no_outer_lock(piLocal);
  }

  for (auto pi : usedByList) {
    Package_sp piLocal = gc::As<Package_sp>(CONS_CAR(pi));
    piLocal->unusePackage_no_inner_lock(pkg);
  }
`````
and your example works now fine.

Will update the pr with the change and a regression-test

kpoeck:
updated the pr as commented

Bike:
yeah, probably the use/used-by lists were always screwed up, but we just ignored it in `delete-package`.

Bike:
alright, i've merged this into my branch and all the waf tests pass now. after i run ansi tests again i'll merge into master

PR #1156, opened 2021-03-31
Silence obj_deallocate_unmanaged_instance

yitzchak:
This printf dumps the whole printable representation of the object which is especially irritating when using static-vectors. Not sure if the message is really needed. If so should it be shorter or maybe a trace message facility that one can turn off? For now I just commented it out to avoid a slew of messages in the cando-jupyter molecule trajectory stuff.


Bike:
i've been in the habit of deleting these bare printfs whenever i find them. they're usually the remainder of a printf debugging session that is no longer relevant.

Issue #1155, opened 2021-03-29
interface for cando


Bike:
Since cando and clasp have grown out of the same project, cando uses clasp "internals" quite generously. It would be more convenient for maintenance if cando used a defined interface to clasp, so that what cando uses can be specially protected during clasp development.

From a quick grep, the cando lisp code has 192 uses of the `core:` package. (The cando C++ code's use of clasp internals will be a whole other kettle of fish.) Those uses are as follows:

- 94 uses of `core:make-cxx-object`. This should probably be exported from a defined Clasp system for C++ interfacing.
- 12 uses of `core:native-vector<int>` related operators. Not sure what's going on with those, but could be an extension.
- 8 uses of `core:bformat`. Some of these are used in logging and this should probably be replaced with `cl:format` if possible. There are also uses in the Fortran printer. I don't know if `cl:format` can handle that, but hopefully it should be adequate?
- 5 uses of `core:split`. This function seems to have the same basic operation of the Lisp split-sequence library (https://github.com/sharplispers/split-sequence), which could be used. `split` is kind of a general utility function so having it as part of clasp's interface would be a little weird, but we could do that too. There is also one use of `core:split-at-white-space`.
- 2 uses of `core:proper-list-p`, which has similar issues.
- 5 uses of `core:encode`. Could probably go with `core:make-cxx-object`?
- 5 uses of `core:getpid`, 3 of `core:num-logical-processors`, 1 of `core:fork`, 1 of `core:fork-redirect`, 1 of `core:sigchld-count`, 1 of `core:wait`, 1 of `core:wifexited`, 1 of `core:wifsignaled`, 1 of `core:wtermsig`, 1 of `core:lseek`, 1 of `core:read-fd`, 1 of `core:close-fd`, 2 of `core:mkstemp-fd`, 7 of various `core:sig` functions, 3 of `core:chmod`. These are all POSIX related functions that we should export from a dedicated POSIX interface.
- 1 use of `core:strftime`, which doesn't actually seem to be defined anywhere, but which would presumably be an interface to the unix function.
- 2 uses each of `core:argv` and `core:argc`, and 1 of `core:*command-line-arguments*`, which should be made into an exported interface.
- 4 uses of `core:quit`. This has already been moved to the `ext:` package, i.e. calls to `core:exit` can and should be replaced with calls to `ext:quit`. There are also 11 uses of `core:exit` and 1 use of `core:c_exit`; these functions may be redundant?
[...]

PR #1159, opened 2021-04-07
Rename core:is-interactive-lisp to core:interactivep and export in :ext

yitzchak:
Export from `:ext` since Cando is using `core:is-interactive-lisp` to determine the verbosity of quickload and progress bars. Renamed to `interactivep` since `is-interactive-lisp` doesn't really follow the lisp predicate style. If `interactivep` is not descriptive enough then maybe `interactive-session-p`?

I searched GitHub for uses of `core:is-interactive-lisp` and found only Cando and Roswell. Roswell doesn't appear to be using it in their own code, it just shows up because they have a Clasp fork.

For #1155

### DIFF
--- a/src/core/lisp.cc
+++ b/src/core/lisp.cc
@@ -1467,7 +1467,7 @@ CL_DEFUN bool core__debugger_disabled_p() {
   return _lisp->_DebuggerDisabled;
 }
 
-CL_DEFUN bool core__is_interactive_lisp() {
+CL_DEFUN bool core__interactivep() {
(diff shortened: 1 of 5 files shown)


Bike:
so the reason i mentioned `is-interactive-lisp` is that it controls more than one thing and could be set by multiple things. It's a little bit messy conceptually. For example we could control verbosity independently. Or, someone could start Clasp with the repl suppressed, but load their own repl with an eval option, thus making the session interactive but with `is-interactive-lisp` false.

PR #1160, opened 2021-04-11
Silence all remaining warnings for unused variables

kpoeck:
I am trying to remove all warnings from cclasp compilation in order to be able to find warnings which are real problems. First batch of changes was  #1157

Doing that I try to just by adding:
*  (declare (ignore ..)) or 
*  (declare (ignorable ))
* Sometimes _ is used twice in a (let* ()), than i disambiguate the variables with `_`, `_1`,..
* In https://github.com/clasp-developers/clasp/blob/master/src/lisp/kernel/clos/sequences.lsp#L967 I believe there was a real error and initial-contents should be used instead of initial-element
* Next step would be to adapt all systems in `contrib`

Testing:
* regression-tests
* ansi-tests
* rebuild 120 most popular quicklisp systems including ironclad and mcclim
* included cando in the build
* executing `icando-boehm`
* executing `ileap-boehm`


Bike:
the sequences code does look genuinely wrong. easy fix.

kpoeck:
when I said next step, I meant a new PR, this one here should be complete

Bike:
things that look like actual bugs:

- sequence:nsubstitute should not ignore start and end. it can probably just pass them to do-general-subsequence
- in defstruct we should need the standard constructor for `#s`. if we do something else now we should just remove the code from the macro.
- if we have `with-tests` that aren't using the `compare` operator we can probably just remove the `with-tests`.
- we can just remove those variables in source-location
- can probably just delete tpl-document-command if it's gonna be that useless

all minor stuff. i can merge this and fix that stuff when i've reached a stopping point with what i'm doing, hopefully today

PR #1170, opened 2021-05-31
Add ed hooks

yitzchak:
Makes `cl:ed` actually do something. Idea copied from SBCL.

### DIFF
--- a/src/lisp/kernel/init.lsp
+++ b/src/lisp/kernel/init.lsp
@@ -188,6 +188,7 @@
 (import 'core:argv :ext)
 (import 'core:rmdir :ext)
 (import 'core:mkstemp :ext)
+(import 'sys::*ed-functions* :ext)
 
 ;;; EXT exports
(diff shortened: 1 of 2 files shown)


yitzchak:
Please note that ccl and allegro also have something similar.  Ecl appears use the environment variable `EDITOR` and cmucl uses hemlock.

Bike:
I'll do a little bit of cleanup and then merge this. Out of curiosity, do you actually use the `ed` function?

yitzchak:
I am exploring using in Jupyter. There is a message that you can pass from the kernel to the frontend UI that will open a file in an editor. It doesn't work in Jupyter Lab, but it does work in the Qt console. Just playing around right now, though.

yitzchak:
Here is a video of it being used in SBCL if you are curious. https://1drv.ms/u/s!Aov9gbLSddWe0HZCBUovvBIRFASF?e=3gMpsK

PR #1175, opened 2021-08-05
Use lambda list in frame-locals

yitzchak:
Attempt parse lambda list using `core:make-lambda-list-handler` and fall back to original method if that fails.

### DIFF
--- a/include/clasp/core/lambdaListHandler.h
+++ b/include/clasp/core/lambdaListHandler.h
@@ -269,7 +269,7 @@ public:
 	  * VALUES(3) = key-flag				; T if &key was supplied
 	  * VALUES(4) = (N key1 target-var1 init1 target-flag1 ... )	; keyword arguments
 	  * VALUES(5) = allow-other-keys			; flag &allow-other-keys
-	  * VALUES(6) = (N target-aux1 init1 ... )		; auxiliary variables
+	  * VALUES(6) = (target-aux1 init1 ... )		; auxiliary variables
 	  *
 	  * 1°) The prefix "N" is an integer value denoting the number of
 	  * variables which are declared within this section of the lambda
--- a/src/lisp/kernel/lsp/debug.lsp
+++ b/src/lisp/kernel/lsp/debug.lsp
@@ -102,32 +102,68 @@ If the arguments are not available, returns NIL NIL."
                    (core:debugger-frame-args-available-p xep))))
         (t ; nothin
          (values nil nil))))
-(defun frame-locals (frame)
-    "Return an alist of local lexical variables and their values at the continuation the frame represents. The CARs are variable names and CDRs their values.
-Multiple bindings with the same name may be returned, as there is no notion of lexical scope in this interface."
-  ;; TODO: This is not a real solution, at all.
-  ;; At minimum, we should use the lambda list's parameter names when available
-  (let* ((fname (frame-function-name frame))
-         (args (frame-arguments frame)))
-    ;; KLUDGE (on top of a kludge) to do better with method arguments
-    ;; We have two kinds of method functions - fast and not - but we
-    ;; name them the same. The former we can treat normally, but the
-    ;; latter have a vaslist of the real arguments as their first, and
-    ;; the next method list as the second.
-    (if (and (consp fname)
-             (eq (first fname) 'cl:method)
-             (= (length args) 2)
-             (core:vaslistp (first args)))
+
+(defun lambda-list-alist (lambda-list arguments eval &aux bindings)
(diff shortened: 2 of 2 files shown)


Bike:
I don't think `names-of-lexical-variables` is the thing to use here. For example if we have `(defvar *test*) (defun foo (*test*) ...)`, the lambda list handler for `(*test*)` will report no lexical variables, which is correct of course, but not for this purpose. I am also not sure how well this works with more complex lambda lists, like ones with `-p` variables.

Bike:
Yeah, there'd be problems with `-p` too. E.g. with a lambda list of `(&optional (x nil x-p) y)`, `names-of-lexical-variables` is `(x x-p y)` but frame arguments is `(xvalue yvalue)`.

yitzchak:
Seems like `core:classified-symbols` symbols includes the special variables.

```lisp
(defvar *test*)
(defparameter fd (core:make-function-description :function-name 'fu :lambda-list '(*test*)))
(core:classified-symbols (core:make-lambda-list-handler (core:function-description-lambda-list fd)))
```
returns `((EXT:SPECIAL-VAR . *TEST*))`

yitzchak:
Also for your example.

```lisp
(defun fu (*test*))
(defparameter fd (core:make-function-description :function-name 'fu :lambda-list '(*test* &optional (x nil x-p))))
(core:classified-symbols (core:make-lambda-list-handler (core:function-description-lambda-list fd)))
```
returns `((EXT:SPECIAL-VAR . *TEST*) (EXT:LEXICAL-VAR X . 0) (EXT:LEXICAL-VAR X-P . 1))`

So am I right in that it handles `-p` fine?

yitzchak:
I think I will have to do the equivalent of `createBindingsInScopeVaList`

yitzchak:
@Bike Ready for another look

Bike:
I think this is a lot closer. Unfortunately I still have nitpicks. First, you can use `constantp` and `ext:constant-form-value` (with no lexenv I guess) to get the value of constant defaults. This will include quotation for example. Second, if I'm reading this code correctly, if the default form is not constant, the variable is reported to the debugger/user as being bound to NIL. Since that isn't generally true, I think maybe we shouldn't report an available binding for such variables - seems better than an incorrect one.

Third we should maybe use Clasp's builtin lambda list parser (`core:process-lambda-list`) rather than Cleavir's, because that means this will work in bclasp and we don't need to worry about the CST stuff. Fourth there should be tests, since this is kind of involved. But I've already nitpicked enough and I can do that later rather than you having to do it.

yitzchak:
I don't have a problem fix that stuff myself. No worries!

yitzchak:
I did have an idea that would increase the complexity a bit, but would make it possible to leave it up to the user whether or not to `eval` the init forms. Basically, we could add an `:eval` key that takes a function like `(eval form alist)` where `alist` is the bindings accumulated so far. By default this key would be `nil`. In that case a variable requiring the evaluation of an init form would just be skipped. If the user provided the `eval` function then the variable would be included with the value being the return of the supplied `eval` function.

In my own debugger I would probably pass an eval function based on some dynamic variable.

Thoughts?

yitzchak:
@Bike I pushed changes with my eval idea. It turned out pretty simple to implement. I also fixed the comment about auxiliary variables in `core:process-lambda-list`.

Bike:
I am pretty skeptical of any kind of evaluation for the reasons i mentioned on IRC, and would much rather we find a way to just get the actual value through DWARF or whatever. However that is not forthcoming and I don't mind having an evaluator as an optional add-on.

So I think there are only two little problems left. One, you should use `ext:constant-form-value`. Let me explain that in a little more detail. Say you have a lambda list `(&optional (x 'foo))`. Then the default form will be `'foo`, aka the list `(quote foo)`. `constantp` will return true on that, since it is a constant. If you just return the form like you do here, the debugger will report X's value as the list `(quote foo)`, rather than what it should be, the symbol `foo`. `constant-form-value` will also resolve constant variables. All you have to do here is replace `(cons var init-form)` with `(cons var (ext:constant-form-value init-form))`; i'd do it myself but I don't want to mess with your force pushes.

The second thing is that I think the `for (indicator value nil) = (multiple-value-list (get-properties arguments (list key)))` is technically illegal. `nil` doesn't mean "ignore", though we might be mistakenly treating it that way. You could do `for (indicator value) = ...` and I think that would work (and the third value would be ignored), or you could just have `do (multiple-value-bind (indicator value) (get-properties ...) (bind-var ...))`.

yitzchak:
Fixed for `ext:constant-form-value`

PR #1177, opened 2021-08-14
Don't dump the library, if a symbol was not found

kpoeck:
* Was not present in master, probably introduced while updating to llvm13
* Gave a lot of undesired output in cffi-test


Bike:
FYI when you make this kind of request you can usually just actually delete them, rather than only comment them out.

kpoeck:
ok, will be deleting in the future

PR #1185, opened 2021-09-19
DO NOT MERGE - Add strict slot option for WSCL type-error conformance 

yitzchak:
This PR adds a `:strict` slot option to make it easy to conform to WSCL issues like [CELL-ERROR-NAME-TYPE-ERROR](https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/cell-error-name-type-error). One could just define a method like the following for each case, but there is 12 such functions for just the condition readers.
```lisp 
(defmethod cell-error-name (instance)
  (error 'type-error :datum instance :expected-type 'cell-error))
```
I haven't thought much about a good name for the option. `:strict` was the first one that came to mind. Could do something like `:strict-accessors` or something. Also, I haven't looked at change.lisp to make sure it is compatible yet.

Could use some feedback on this PR!

### DIFF
--- a/src/lisp/kernel/clos/conditions.lsp
+++ b/src/lisp/kernel/clos/conditions.lsp
@@ -522,11 +522,15 @@ returns with NIL."
 (define-condition error (serious-condition) ())
 
 (define-condition simple-condition ()
-  ((format-control :INITARG :FORMAT-CONTROL :INITFORM ""
-		   :ACCESSOR simple-condition-format-control)
-   (format-arguments :INITARG :FORMAT-ARGUMENTS :INITFORM NIL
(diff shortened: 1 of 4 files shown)


Bike:
interesting idea, but it seems like a lot of work, and we could just use macros to shorten the number of defmethods needed.

if we were to do something like this, i'm pretty sure we should use `core:strict` or something instead of a keyword, because otherwise we'd potentially interfere with user extensions.

yitzchak:
Yeah, it was a brainstorm and I was kind of on the fence regarding its usefulness. I'll keep the branch around just in case and just do a new PR with the methods directly defined.

PR #1186, opened 2021-09-24
Add methods to signal type errors for WSCL conformance

yitzchak:
I did not add `setf` methods for the `simple-condition` slots since the spec doesn't specify that `setf` methods are available. I don't see any use of those methods in CLASP so I am wondering if that is a typo.


Bike:
They don't have writers. Condition slots are basically immutable.

PR #1190, opened 2021-10-02
Backport fix for esrap for probelms with newest sbcl

kpoeck:
Problem, the definition:
```lisp
(declaim (ftype (function * (values function &optional))
                resolve-function))
````
does not fit with the definition of `resolve-function` that returns the result of `compile` which returns 3 values.

Newest sbcl gives an error with that.


Bike:
Can we not just update esrap?

kpoeck:
Can also try that, but the version in quicklisp is not yet updated, neither is there a tag nor a release of esrap.

Bike:
yeah, got that covered.

Bike:
see: 38aae2bc1808f846161020da2b051b1e36b6aa97

PR #1201, opened 2021-10-24
Get xcode_sdk from wscript.config

kpoeck:
XCODE_SDK was hardcoded.
Get it from wscript.config with the old value being the default.
LLVM_HASH is not used anymore


kpoeck:
something happened to this or, can no longer update (closing)

PR #1209, opened 2021-10-30
Cleanup exports in :core

kpoeck:
* `core__current_dir` is declared but never used
* lots of symbols are exported but never used
* some symbols are used in c++ land but never in lisp, so don't need to be exported, just defined
* `(list*)`calls `FEargument_number_error`which raises `core::_sym_argument_number_error` which is not defined in lisp. Defined  `core:argument-number-error` as a subclass of `core:wrong-number-of-arguments`mapping the initargs.
* in src/lisp/kernel/cmp/primitives.lsp protected `%long-float%` with `#+long-float (:long-float %long-float%)`gave an undefined function warning.
* `queue-p` is really  `queuep`
* run regressions-tests ansi-tests i-cando-boehm and lots of quicklisp systems


Bike:
We don't need a new argument number error subclass for `list*` just because drmeister or I misremembered the condition type name or failed to upgrade it or something. We actually don't even need `FEargument_number_error` at all, since it's only called in one place, and there's already `throwTooFewArguments` to accomplish the same task.

PR #1213, opened 2021-11-06
Cleanup package cmp

kpoeck:
as with the other cleanups
* don't export symbols that don't need o be expored
   * do that in c++
   * do hat in lisp
   * also conditionalize some variables, that need to be e.g. `#+tag-bits4 +fixnum10-tag+`
* test with
   * regression tests
   * cando
   * ansi-tests
   * some quicklisp packages


Bike:
is `tag-bits4` defined as a feature anywhere?

kpoeck:
yes in https://github.com/clasp-developers/clasp/blob/main/src/core/lisp.cc#L1268

PR #1230, opened 2021-12-23
Allow real-valued columns in Gray stream methods

yitzchak:
`pprint-indent` allows real column values not just integers. Most print printers assume integers though since they were built for monospaced fonts. The Gray stream methods are ambiguous about the allowed values but should deal with real values to make `stream-advance-to-column` work for a proportional pretty printer.

Corresponding ECL PR: https://gitlab.com/embeddable-common-lisp/ecl/-/merge_requests/264
McCLIM PR that is currently reverted until issue is resolved in important implementations: McCLIM/McCLIM#1248
Some discussion here: https://irclog.tymoon.eu/libera/%23clim?around=1639832373#1639832373

FYI @dkochmanski


Bike:
huh, interesting. thank you, merged

PR #1241, opened 2022-01-09
Don't assume that loop variables for <in> and <on> really hold lists

### LINKED ISSUE #1212: (loop for x in '(a . b) collect x) -> SEGMENTATION-VIOLATION
```lisp
Starting cclasp-boehmprecise-0.4.4-979-g01e77db41-cst ... loading image...
Writing jitted symbols to /tmp/perf-3443.map
Entered cando-startup.lisp
Starting start-cando.lisp script - arguments: #()
Done load start-cando.lisp in cando-startup.lisp
To load "cando-user":
  Load 1 ASDF system:
    cando-user
; Loading "cando-user"
...........
Top level in: #<PROCESS TOP-LEVEL @0x10fd8b421 (Running)>.
COMMON-LISP-USER> (loop for x in '(a . b) collect x)

Condition of type: SEGMENTATION-VIOLATION
Segmentation fault. Attempted to access resticted memory address #x0.

Available restarts:

../../src/core/backtrace.cc:154:args_from_offset When trying to get arguments from CL frame read what should be a closure 0x1ea8f but it isn't

(EXT:SEGMENTATION-VIOLATION 0)
````

### LINKED ISSUE #1239: Bad LOOP macro segfaults
```
(defun screw-up ()
  (loop for x in 42
        do (format t "x = ~a~%" x)))
```
Will segfault.  Nothing should segfault.

kpoeck:
* fixes #1212 
* fixes #1239
* by
   * don't use core:cons-car since it assume it argument to be cons
   * don't use core:cons-cdr since it assume it argument to be cons
   * don't declare to variable for <in> or <on> to be a list


Bike:
as i mentioned in #1239, the declaration is useful, since for example the code for dealing with null lists can be deleted even if the safety type check is not.

kpoeck:
this pr no longer merges

PR #1247, opened 2022-01-17
Put neutral default for XCODE_SDK_DEFAULT that doesn't use the version number of macos

kpoeck:
and drop unused LLVM_HASH


Bike:
I don't really build on OS X right now, so I don't have much ability to sanity check this (I'm sure it's fine, though)

PR #1256, opened 2022-01-24
Fix another branch of v for types that are symbols + test

kpoeck:
* forgot to fix another branch also relevant to compiled mcclim
* added regression-test


kpoeck:
@Bike, could you check when you have time

PR #1271, opened 2022-03-05
More details for apropos

kpoeck:
Extension of apropos:
* (unbound) global vars
* class
* Type


Bike:
Why is there a colon on the unbound variable notice? There's nothing after that, is there?

kpoeck:
Seems to be an error, so will drop that

kpoeck:
@Bike fixed

PR #1272, opened 2022-03-05
Declare types better or more correct to avoid - correct - compiler warnings

kpoeck:


Bike:
in what situation would an element of `*elementary-types*` or `*member-types*` be nil? It doesn't seem possible from the code in predlib.

kpoeck:
if you macroexpand the code  e.g. for elementary types you get
```lisp
(BLOCK NIL
  (LET* ((%DOLIST-VAR *ELEMENTARY-TYPES*) I)
    (DECLARE (TYPE CONS I))
    (WHILE %DOLIST-VAR
           (SETQ I (FIRST %DOLIST-VAR))
           (LET ((OTHER-TYPE (CAR I)) (OTHER-TAG (CDR I)))
             (WHEN (FUNCALL IN-OUR-FAMILY-P OTHER-TYPE)
               (COND
                ((FUNCALL TYPE-<= TYPE OTHER-TYPE)
                 (IF MINIMIZE-SUPER
                     (WHEN (ZEROP (LOGANDC2 OTHER-TAG SUPERTYPE-TAG))
                       (SETQ SUPERTYPE-TAG OTHER-TAG))
                     (SETQ SUPERTYPE-TAG (LOGIOR OTHER-TAG SUPERTYPE-TAG))))
                ((FUNCALL TYPE-<= OTHER-TYPE TYPE)
                 (SETQ SUBTYPE-TAG (LOGIOR OTHER-TAG SUBTYPE-TAG)))
                (T (SETQ DISJOINT-TAG (LOGIOR DISJOINT-TAG OTHER-TAG))))))
           (SETQ %DOLIST-VAR (CDR %DOLIST-VAR)))
    NIL))
````

Bike:
I see. I just checked the CLHS, and it's pretty clear that declarations do apply to the result form, and that the variable will be nil at that time. Well that's a little dopey. Oh well.

PR #1276, opened 2022-03-10
Rename all .lsp files to .lisp

yitzchak:
I updated comments that mentioned `.lsp` files also. Some of the documentation files are probably out of date and need more attention, but I updated the extensions anyways.

Built on arch without extensions once just make sure I didn't break anything obvious.


Bike:
lemme just make sure it builds with cando and then i'll push the merge

PR #1291, opened 2022-04-13
Don't use --non-interactive on regression tests

yitzchak:


kpoeck:
I observe the same failures w/o --non-interactive, perhaps this is not needed

yitzchak:
Yeah. Something else is going on.

PR #1312, opened 2022-05-17
Fix sxhash for symbols (see ansi SXHASH.15 SXHASH.23)

kpoeck:
Fix ansi
* SXHASH.15
* SXHASH.23

using package information to compute the hash is incorrect, see using package information to compute the hash


Bike:
Wow, that's all it was? Nice job working that out

PR #1314, opened 2022-05-17
more informative error message

kpoeck:
* more informative error message for create_single_dispatch_generic_function

### DIFF
--- a/include/clasp/core/singleDispatchGenericFunction.h
+++ b/include/clasp/core/singleDispatchGenericFunction.h
@@ -108,7 +108,7 @@ public:
         }
       }
     }
-    SIMPLE_ERROR(("This single dispatch generic function %s does not recognize argument class %s")
+    SIMPLE_ERROR(("This single dispatch generic function %s does not recognize argument class %s  arg %s")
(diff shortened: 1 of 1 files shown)


Bike:
It looks like the format string has two `%s`s but three arguments. That's not right, is it?

kpoeck:
You are obviously right, fixed that.

Test e.g. with `(LLVM-SYS:GET-VENDOR-NAME 23)`
results in
```lisp
Condition of type: SIMPLE-PROGRAM-ERROR
This single dispatch generic function #<SINGLE-DISPATCH-GENERIC-FUNCTION GET-VENDOR-NAME> does not recognize argument class #S(BUILT-IN-CLASS FIXNUM )  arg 23
Available restarts:

(CORE:SIGNAL-SIMPLE-ERROR CORE:SIMPLE-PROGRAM-ERROR NIL "~a" ("This single dispatch generic function #<SINGLE-DISPATCH-GENERIC-FUNCTION GET-VENDOR-NAME> does not recognize argument class #S(BUILT-IN-CLASS FIXNUM )  arg 23"))
````

PR #1317, opened 2022-05-18
FD_SET FD_ISSET FD_CLR work again on darwin

kpoeck:
* undo FD_SET causes problems with Xcode 11.4 - remove this error message once it's working again


drmeister:
Thank you!

PR #1325, opened 2022-05-25
Improve static analyzer

yitzchak:
- Move parallel analyzer logs into build directory
- Put all analysis results into src/analysis
- Add bash script to run analyzer on all significant
  extension combinations


yitzchak:
Before this can be merged the corresponding branch on cando has to be merged and the branch reference in `repos.sexp` has to be changed back to `cando/main`.

drmeister:
I think I see - I should change this to point to cando/main: `https://github.com/clasp-developers/clasp/blob/analyzer/repos.sexp`
And merge cando/analyzer into cando/main

yitzchak:
Correct

PR #1328, opened 2022-05-27
Fix snapshot command line argument handling

yitzchak:
Some of the command line options are moved into `globals_t` during Lisp startup. This code is not executed for snapshots so these command line arguments will not work. For example `clasp --quit` works but `dclasp --quit` does not. The affected options are: `--ignore-image`, `--noinit`, `--noinform`, `--noprint`, `--quit`, `--norc`, `--rc`, `--snapshot-symbols`, and `--disable-debugger` plus meta-options that depend on these options such as `--non-interactive`. Obviously, some of these options have nothing to do with snapshots, while some do.

To fix this I have removed the duplicate state contained in `core::globals_` and used `core::global_options` exclusively to store the command line option state.

Also, made a few tweaks to raw argument processing.


yitzchak:
Just realized that I'll need to initialize some of the lisp variables related to command line options in the snapshot. So a bit more work to do.

PR #1331, opened 2022-06-05
Regression tests for wrong parameters for room

kpoeck:
from clhs:
Arguments and Values:
* x - one of `t`, `nil` or `:default`.


Bike:
should we more specifically require a type error?

kpoeck:
and the type would be (or t nil :default)?

The code doesn't do that, but I obviously can change that too.

Should I change it?

kpoeck:
I will change to test to type-error and change clasp to signal a type-error

Bike:
I don't see why not. `room` isn't exactly a hot loop function where we need to concern ourselves with performance

kpoeck:
@Bike did as you proposed, so please check

PR #1340, opened 2022-06-11
Allow dolist with declarations for the loopvar

kpoeck:
Fix dolist to that the loopvar can be declared as having a type
* allows to further cross-compile sbcl with clasp (fails later anyhow)
explanation:
```lisp
(macroexpand '(dolist (a (list 1 2 3) sum)
                     (declare (type fixnum a))
                     (incf sum a)))
````
used to expand to
```lisp
(BLOCK NIL
  (LET* ((%DOLIST-VAR (LIST 1 2 3)) A)
    (DECLARE (TYPE FIXNUM A))
    (WHILE %DOLIST-VAR
           (SETQ A (FIRST %DOLIST-VAR))
           (INCF SUM A)
           (SETQ %DOLIST-VAR (CDR %DOLIST-VAR)))
    (SETQ A NIL)
    SUM))
`````

which obviously is wrong, since a is initialized to nil and later declared as fixnum    
    
with the fix:
```lisp
(BLOCK NIL
  (LET* ((%DOLIST-VAR (LIST 1 2 3)))
    (WHILE %DOLIST-VAR
           (LET ((A (FIRST %DOLIST-VAR)))
             (DECLARE (TYPE FIXNUM A))
             (TAGBODY (INCF SUM A) (SETQ %DOLIST-VAR (CDR %DOLIST-VAR))))))
  (LET ((A NIL))
    (DECLARE (IGNORABLE A))
    SUM))
````
coments:
* A is allowed to be locally definde in the `while``
* Special care needs to be applied, so that
   * A finally is bound to nil
   * tagbodies are still respected
   * parat from regression tests and ansi tests I tested over 100 quicklisp systes with the change

### DIFF
--- a/src/lisp/kernel/lsp/export.lisp
+++ b/src/lisp/kernel/lsp/export.lisp
@@ -57,15 +57,18 @@
                (multiple-value-bind (declarations body)
                    (process-declarations body nil)
                  `(block nil
-                    (let* ((%dolist-var ,expr)
-                           ,var)
-                      (declare ,@declarations)
+                    (let* ((%dolist-var ,expr))
                       (si::while %dolist-var
-                                 (setq ,var (first %dolist-var))
-                                 ,@body
-                                 (setq %dolist-var (cdr %dolist-var)))
-                      ,(when exit `(setq ,var nil))
-                      ,@exit)))))))
+                        (let ((,var (first %dolist-var)))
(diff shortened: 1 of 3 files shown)


kpoeck:
I forgot to add that export.lisp is a totally wrong name, but I agreed to change that after the bigmerge for llvm14

PR #1343, opened 2022-06-15
Add eclasp image

yitzchak:
Adds an extra stage when extensions are included in the build. a -> b -> c -> e [-> s]. 

In order to control when extensions are loaded I've had to add another feature `:IGNORE-EXTENSION-SYSTEMS`. When the selected stage is "c" then extensions will be loaded by default. If the feature `:IGNORE-EXTENSIONS` is present then no extensions will be loaded and the system will behave like stock Clasp. If the feature `:IGNORE-EXTENSION-SYSTEMS` is present then the `CL_INITIALIZER` functions will still initialize their extension code, but the various extension systems in `*EXTENSION-SYSTEMS*` will not be loaded with `QL:QUICKLOAD`/`ASDF:LOAD-SYSTEM`. This is needed during compilation of the eclasp image.

I've rewritten the ASDF groveler to __NOT__ use ASDF's  perform methods. It turns out to be virtually impossible to trick ASDF into using a different value of `*FEATURES*` for introspection of an ASDF system versus compilation/loading of the same system. Several of the systems in Cando have feaure expressions in their ASD files. To address this I've made pull requests to those repos and changed the read time conditionals into ASDF style `:feature`/`:if-feature` expressions.

I've also added Cando jobs to the testing workflow along with Cando regression tests via `ninja -C build cando-test`


kpoeck:
I have another pr 1342 hat also modifies clasp-builder. Please don't merge this one, before mine is through.

kpoeck:
I see you also find the incorrect #+(not bclasp cclasp) for cmp::with-compiler-timer. This is defined in compile-file.lisp. Now I see compile-file is part of aclasp, so that should be ok

yitzchak:
Yeah, it showed up as an error in compilation of the eclasp image, presumably because the full compiler is available there.

PR #1344, opened 2022-06-15
 Fix some issues with source info

yitzchak:
- Use SYS logical host for source info pathnames. There are some non-permitted characters such as underscores, but they don't seem to cause an error.
- Transfer docstrings for wrapped functions
- Add SETF support to scraper
- Improve SETF support in DESCRIBE and DOCUMENTATION
- Add regression tests for scaper docstrings and SETF functions


Bike (approved):
We might want to think about having a `setf_method_kind`, but we don't need it right now. Looks good.

PR #1349, opened 2022-06-23
Use logical paths in source references

yitzchak:
This PR adds support for logical pathnames in source info references. This is accomplished via the following:
1, Modules - adding `:source-debug-pathname` during compilation
2. Images - Update clasp-builder to use all logical pathnames and avoid "searching" for the location of source files and instead rely on `translate-logical-pathname`.
3. Scraper - Pathnames extracted by the scraper are converted to logical pathnames under the SYS host and added to the translation tables as needed by `define_pathname_translation`.

The main difficulty with logical pathnames is of course the limited characters available for each component (uppercase alphanumeric characters/hyphens). We have case sensitive pathnames and pathnames with underscores, therefore the translation is not straightforward. To address this each source pathname is checked for characters that prevent a trivial translation. A trivial translation is one in which each pathname component is composed of only lowercase letters, numbers and hyphens.

If a pathname does not permit a trivial translation then it is reduced to the "shortest" non-trivial pathname needed. For example, `src/fu_bar/Wibble/baz/quux.c` is reduced to `src/fu_bar/Wibble/**/*.*`, whereas `src/fu_bar/Wibble/baz/Gronk.c` is already shortest because the name component has uppercase letters.

Once a pathname is reduced to the shortest non-trivial pathname needed for translation then it is converted to a logical pathname by replacing all non-alphanumeric letters with hyphens and setting the version component to `:wild`. This is the pathname used in the "from" mapping in `logical-pathname-translations`. For the examples above this results in the following when the SYS root is `/usr/local/clasp/`:

```lisp
(logical-pathname-translation "SYS") ; =>
((#P"SYS:SRC;FU-BAR;WIBBLE;**;*.*.*" 
  #P"/usr/local/clasp/src/fu_bar/Wibble/**/*.*")
 (#P"SYS:SRC;FU-BAR;WIBBLE;BAZ;GRONK.C" 
  #P"/usr/local/clasp/src/fu_bar/Wibble/baz/Gronk.c")
 (#P"SYS:**;*.*.*" 
  #P"/usr/local/clasp/**/*.*")
```


Bike (commented):
in the release notes: "asynchronous"
 i don't quite follow how the GENERATED and FASL logical hosts are used in koga - those aren't something we use in the built clasp, are they? since the only host we have reserved is SYS
does this work in gdb/lldb? i'm building now and will take a look myself

yitzchak:
There are a number of logical hosts set up in bundle.cc. GENERATED, FASL, LIB, BITCODE and SYS are the really important ones.  FASL, LIB and BITCODE are currently synonyms for each other. Probably historical. During build the important hosts are
- FASL = build/boehmprecise/fasl/
- SYS = .
- GENERATED = build/boehmprecise/generated/

After install
- FASL = /usr/lib/clasp/
- SYS = /usr/share/clasp/
- GENERATED = /usr/share/clasp/generated/

Bike (approved):
hardcoding /tmp in the unwind regression tests might cause a failure on some systems. but that's a pretty minor thing.

yitzchak:
Maybe we should do something like UIOP? https://gitlab.common-lisp.net/asdf/asdf/-/blob/master/uiop/stream.lisp#L544

Seems kind of silly to make a translation for `SYS:TMP;**;*.*.*` .... maybe just a function in core?

PR #1354, opened 2022-07-04
Generate alpha-char-p, alphanumericp and graphic-char-p from Unicode data

yitzchak:


yitzchak:
@kpoeck I have found a detail that we need to resolve in the logic for GRAPHIC-CHAR-P. Specifically, the spec says

> graphic adj. (of a character) being a "printing'' or "displayable'' character that has a standard visual representation as a single glyph, such as A or * or =. Space is defined to be graphic. Of the standard characters, all but newline are graphic. See non-graphic. 

In extending to Unicode this seems to me to be the [Unicode categories](https://en.wikipedia.org/wiki/Unicode_character_property#General_Category): Ll, Lm, Lo, Lt, Lu, Mc, Me, Mn, Nd, Nl, No, Pc, Pd, Pe, Pf, Pi, Po, Ps, Sc, Sk, Sm, So, and Zs. Note that Space resides in Zs.

There is also a definition in the spec glossary for "printing."

> printing adj. (of a character) being a graphic character other than space. 

This is used in the implementation of `~:C` in FORMAT which spells out non-printing characters while using WRITE-CHAR for others. Our current implementation uses this https://github.com/clasp-developers/clasp/blob/da3906a15dfd99aa7d5ce0919c87c129106b5ad0/src/lisp/kernel/lsp/format.lisp#L882-L891

If we include Zs in GRAPHIC-CHAR-P then there will space-like characters, other than Space itself, that will be printed with WRITE-CHAR when using ~:C` in FORMAT.

If we interpret the definition of "printing" to be

> printing adj. (of a character) being a graphic character other than space-like characters.

then CHAR-PRINTING-P could conceptually be GRAPHIC-CHAR-P - Zs, which would cause WRITE-CHAR to be used for all categories except Zs. This seems to me to be in the "spirit" of the spec.

Thoughts?

@Bike language lawyers welcome.

Bike:
I don't quite understand why the spec is making the distinctions it does. Notably, some other whitespace characters like tab and newlines are explicitly _not_ graphic (13.1.4.1). But in Unicode these are not in Zs anyway. It might be worth trying to see if any other operator in the standard makes a distinction between printing and non-printing characters. I don't know if there are any. There's no standard `printing-char-p`.

I think practically speaking, if someone's using `~:c` they probably want the character name printed if the character is not obviously visible. Spaces and newlines are both not visible in this way, and looking at Unicode none of the Zs characters really are either.

Also not sure how `graphic-char-p` is used. I checked McCLIM since it's big, and as you can see here <https://github.com/McCLIM/McCLIM/search?q=graphic-char-p> a lot of the uses actually seem to be doing `printing-char-p` instead, which is not something we can really help with.

yitzchak:
I looked in the TeX sources for `\term{printing}` and the only references are in [22.3.1.1](http://clhs.lisp.se/Body/22_caa.htm). I can't find any references to actual usage of GRAPHIC-CHAR-P in the spec. Just cross-references, etc. 

There are very little references to the reasoning behind these character sets in http://clhs.lisp.se/Issues/iss026_w.htm

I did find this interesting note in the TeX comments of dict-symbols

```tex
\issue{CHARACTER-PROPOSAL:2-6-2}
%% I replaced this on suggestion of KAB.  It doesn't match CHARACTER-PROPOSAL 2.6.2,
%% so who knows how it got here. -kmp 21-Jan-92
% The \term{characters} in a \term{symbol}'s \term{name} must be \term{graphic},
% and may come from any supported \term{character} \term{repertoire}.
A \term{symbol} may have any \term{character} in its \term{name}.
```

This suggests to me that GRAPHIC-CHAR-P might have originally been intended as test a to see if token a token was a symbol. SInce they relaxed the character requirements of symbols this test wasn't needed anymore.

Bike:
okay, so probably not too important either way. figures. I think making Zs graphic but also space for printing-ness purposes as you said should be fine. The rest of the stuff in Zs looks conceptually similar to space, in that it is blank but has some kind of defined width (e.g. all the this-much-of-an-em spaces). and that's in contrast to e.g. tab characters which space things in different ways depending on context.

kpoeck:
I do't think I have any think to add here

yitzchak:
Ok, well thanks @kpoeck. I reverted the Zs inclusion and added only space just to see if the ANSI tests pass. There was one error I was getting before with Zs included. I'll try to figure out how to make it work with Zs in GRAPHIC-CHAR-P

Bike (approved):

PR #1358, opened 2022-07-14
Compiler improvements

Bike:
Whole grab bag of improvements on the Lisp side of the compiler. Some details in the release notes.

Needs benchmarking.


Bike:
I ran the benchmarks. I think switching to a more SBCL-like type check policy slows some code down as the compiler inserts type checks where it didn't previously, e.g. the fairly useless INTEGER types in the ACKERMANN benchmark. I have some ideas for skipping those but they would take some much more involved compiling. On the other hand, this makes things unambiguously much faster for math heavy code, with for example MANDELBROT/DFLOAT dropping from 1.2s to 30ms.

PR #1359, opened 2022-07-24
Use print-object even when *print-readable* is nil

yitzchak:
The spec states that [`print-object`](http://www.lispworks.com/documentation/lw50/CLHS/Body/f_pr_obj.htm) must obey `*print-readably*` but `__write__` seems to prevent `PRINT-OBJECT` from being called when `*print-readably*` is NIL versus using `__repr__` as a fallback.

The current implementation of `__write__` makes it impossible to define `PRINT-OBJECT` methods for clbind extensions like seqan-clasp.

### DIFF
--- a/src/core/object.cc
+++ b/src/core/object.cc
@@ -509,14 +509,12 @@ void General_O::describe(T_sp stream) {
 }
 
 void General_O::__write__(T_sp strm) const {
-  if (clasp_print_readably() && this->fieldsp()) {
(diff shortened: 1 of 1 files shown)


Bike (approved):

Bike:
Specific language lawyery citation: 22.1.2 combined with 22.1.3.13 says print-object is used in all circumstances except `*print-pretty*` being true.

PR #1374, opened 2022-10-25
Revert longjmp optimizations

yitzchak:
Add PR to confirm git bisect observations and run CI

### DIFF
--- a/include/clasp/core/unwind.h
+++ b/include/clasp/core/unwind.h
@@ -119,22 +119,6 @@ public:
   virtual bool unwound_dynenv_p() { return false; }
 };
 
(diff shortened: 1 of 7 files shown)


karlosz:
Just increase the stack size used for dynenvs. Mac probably shoves more things into the jmp_buf objects than Linux so we need to preallocate more space for the dynenv objects. Should be as easy as changing 256k to 512k

karlosz:
That's quite surprising. You are changing the number of dynenv stack words in the VM struct right?

yitzchak:
Correct. 1M didn't work either.

PR #1380, opened 2022-11-03
Tweaks for clbind and koga

yitzchak:
- Add noAutoExport and setf policy
- Add frameworks for Darwin
- Remove unused function declaration functions

The first commit has the changes. Second commit is just clang-format.

### DIFF
--- a/include/clasp/clbind/class.h
+++ b/include/clasp/clbind/class.h
@@ -4,14 +4,14 @@
 
 /*
 Copyright (c) 2014, Christian E. Schafmeister
- 
+
 CLASP is free software; you can redistribute it and/or
 modify it under the terms of the GNU Library General Public
 License as published by the Free Software Foundation; either
 version 2 of the License, or (at your option) any later version.
- 
+
 See directory 'clasp/licenses' for full details.
- 
+
 The above copyright notice and this permission notice shall be included in
 all copies or substantial portions of the Software.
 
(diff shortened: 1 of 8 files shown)


Bike (commented):
Why the containers.h change? I don't understand how this is related?

Also, could you possibly explain what these policies do - probably not in the code, but here on github or on irc, so i can update the clbind docs. I would guess that the setf policy makes it define `(setf foo)` instead of `foo`, and auto export controls whether the symbol is exported automatically? (And if that happens, some clarification on that would be nice, since our examples do `def` and also `SYMBOL_EXPORT_SC_`)

yitzchak:
The containers.h change was needed for lila. I can split it off if you want. It isn't really needed for this stuff. And yes, you are correct about both of those policies. 

`def` automatically exports the symbols, but sometimes it is nice not to. For example, if you are are creating a low-level function that you are going to create some wrapper code in your system for. The setf only works on DEFUN, not the single dispatch `def` on a class. So there it is convenient to do the following:

```c++
class_<DoubleVector>(s, "double-vector")
    .def("setf_vref", &DoubleVector::setf_vref, noAutoExport(), "Set a specific component of a vector."_docstring);
```

And then create the setf function in lisp:

```lisp
(defun (setf vref) (new-value object index)
  (setf-vref object new-value index))
```

Bike (approved):

PR #1381, opened 2022-11-07
Axe mpl

Bike:
Removes boost::mpl dependency.

This slightly alters the interface of clbind `class_` and `derivable_class_`, in that you have to specify any base classes in the second template parameter. Previously mpl was used to make it possible to do this with any of three of the parameters, for reasons I don't understand. But I couldn't find anything using it that way and it wasn't documented.

Probably shouldn't merge this until we unbreak the tests. Also I'll need to double check that the example clbind uses still work.


Bike:
[The cando thing](https://github.com/cando-developers/cando/commit/f6bdd447e9a76a0efd5d5a817f3696ddfc64dfe0) was unrelated so I'm merging now.

PR #1386, opened 2022-11-11
Update test workflow for Node.js 16

Bike:
WIP: We should wait for the setup-miniconda 2.2 release, whenever that may be (it's currently five months late)


Bike:
2.2 is out

yitzchak (approved):
Looks good!

PR #1387, opened 2022-11-14
Better warning for duplicate CASE keys

Bike:


Bike:
PR to run the tests, since I saw some weird possibly unrelated problems and want to double check things.

PR #1389, opened 2022-11-15
Move print-object calls to write-object

yitzchak:


src/lisp/kernel/clos/print.lisp
```diff
@@ -280,6 +280,23 @@ printer and we should rather use MAKE-LOAD-FORM."
...
+         (write-char #\` stream)
+         (core:write-object (second l) stream))
+        (eclector.reader:unquote
+         (write-char #\, stream)
+         (core:write-object (second l) stream))
+        (eclector.reader:unquote-splicing
+         (write-string ",@" stream)
+         (core:write-object (second l) stream))
+        (otherwise
+         (call-next-method)))
```
Bike, line 297: Is it okay for this method to be on `list`? If we're really using `print-object` pervasively, would there not be methods on `cons` and `null` that take priority? Presumably we don't actually have such methods, if this works, but that's the organization I'd imagine

src/lisp/kernel/clos/print.lisp
```diff
@@ -332,47 +349,15 @@ printer and we should rather use MAKE-LOAD-FORM."
 ;;; ----------------------------------------------------------------------
 
 (defmethod describe-object ((obj t) (stream t))
```
Bike, line 351: Might be better to have this method be on `standard-object` or something instead. We probably have methods for built in classes already, but if not things could get a little weird using `class-slots` and all

src/core/object.cc
```diff
@@ -492,11 +492,7 @@ CL_DEFUN bool core__sl_boundp(T_sp obj) {
 void General_O::describe(T_sp stream) { clasp_write_string(this->__repr__(), stream); }
 
 void General_O::__write__(T_sp strm) const {
-  if (_sym_STARliteral_print_objectSTAR->symbolValue().notnilp()) {
```
Bike, line 495: as discussed on IRC - if you don't mind nuking this "literal print object" thing too, go ahead and do that

src/lisp/kernel/clos/print.lisp
yitzchak: Sounds good. BTW, we don't really have any describe-object methods right now. Most all is handled in the pre-CLOS inspect.lisp

Bike (approved):

PR #1395, opened 2022-11-20
Add ansi-test-subset build command

yitzchak:
Usage:

```
ANSI_TEST_SUBSET=FORMAT.C.2A,FORMATTER.C.2A ninja -C build ansi-test-subset
```


yitzchak:
@Bike can you try this out? It runs the named tests with `:catch-errors` set to nil so that one can inspect the stack on errors.

Bike (approved):

PR #1399, opened 2022-11-30
Cleanup

Bike:
Mostly enables a lot more warnings in Clang.

Draft until I see any warnings on Mac and/or Cando are fixed.


Bike:
Still a lot of warnings. Some macro redefinitions and a lot of reorder-ctor in cando.

PR #1403, opened 2022-12-19
Update print-unreadable-object

yitzchak:
There several places in Cando that do this kind of thing:

```lisp
(defmethod print-object ((o fu:bar) s)
  (print-unreadable-object (o s :identity t)
    (write-string "FU:BAR" s))
  o)
```

The current PRINT-UNREADABLE-OBJECT is in C and doesn't really respect the various print dynamic variables or the current package when printing the type. Also, SBCL uses PPRINT-LOGICAL-BLOCK to print the #< ,,, > block when pretty printing, which is kind of cool. To resolve both of these issues I've moved the guts of PRINT-UNREADABLE-OBJECT into Lisp and now methods like the above that exist just to print a qualified type name aren't needed.

# Changes
- properly print qualified type names
- use pprint-logical-block when pretty printing

### DIFF
--- a/src/lisp/kernel/lsp/iolib.lisp
+++ b/src/lisp/kernel/lsp/iolib.lisp
@@ -302,10 +302,33 @@ the one defined in the ANSI standard. *print-base* is 10, *print-array* is t,
 	   (si:cons-cdr %progv-list)
 	 ,@body))))
 
-(defmacro print-unreadable-object
-	  ((object stream &key type identity) &body body)
-  (if body
-      `(flet ((.print-unreadable-object-body. () ,@body))
-	 (print-unreadable-object-function
-	   ,object ,stream ,type ,identity #'.print-unreadable-object-body.))
-    `(print-unreadable-object-function ,object ,stream ,type ,identity nil)))
+(defun print-unreadable-object-contents (object stream type identity body)
+  (when type
+    (write (type-of object) :stream stream :circle nil
+                            :level nil :length nil)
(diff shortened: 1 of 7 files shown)


src/lisp/kernel/lsp/iolib.lisp
```diff
@@ -302,10 +302,32 @@ the one defined in the ANSI standard. *print-base* is 10, *print-array* is t,
...
+    (funcall body))
+  (when identity
+    (when (or body (not type))
+      (write-char #\space stream))
+    (write-char #\@ stream)
+    (core:write-addr object stream)))
+
+;;; The guts of print-unreadable-object, inspired by SBCL.
+;;; pprint-logical-block isn't available yet, so this function
+;;; will be redefined later.
```
Bike, line 320: Could you note that specifically it's redefined in pprint.lisp? And possibly indicate in pprint.lisp that it is a redefinition. I try to make things really clear when we have to jump through bootstrapping hoops like this.

yitzchak: Of course!

Bike (approved):
Avoiding the double space when no body is provided is a nice touch!

yitzchak:
Thanks! I'd love to get rid of the space if there is no body and only identity or type, but the spec doesn't seem to allow for it to be optional and ANSI test print-unreadable-object.3 tests for `"#< "` when type is NIL, identity is T and no body is present.

PR #1405, opened 2022-12-27
Newer eclector

Bike:
Changes to use the (probable) next Eclector branch. This includes both changes for Eclector master, and changes for the new sharpsign equals protocol that has not yet been merged into Eclector master.


Bike:
Waiting for the upcoming Eclector release before this is merged.

PR #1407, opened 2023-01-03
Cleanup some print-unreadable-object uses

yitzchak:
Fix the following:
1. PRINT-UNREADABLE-OBJECT allows an output stream designator in the spec. This works fine until `:identity t` is used because CORE:WRITE-ADDR doesn't allow a designator.
2. PRINT-OBJECT for classes inserts a "The " before the type unlike any other PRINT-OBJECT method.


Bike (approved):
I don't think extending `write-addr` to take designators is bad, but we should may have `print-unreadable-object` do the coercion too, so that we don't need to do it as many times? But that's p minor

yitzchak:
I added the output stream coercion for the non-pretty print branch, but not to the pretty print branch because PPRINT-LOGICAL-BLOCK does that already.

PR #1408, opened 2023-01-03
Set column to zero while in REPL

### LINKED ISSUE #1348: FORMAT function fails on tabulation.
**Describe the bug**
When using the `format` function on a list, the tabulation breaks.

**Expected behavior**
`(defun print-list(ls) (format t "~{~a:~10t~a~%~}" ls))`
When using this to output a list like`(list :a 1 :b 2 :c 3)`
It should give me the following, which is the output of clisp:
```
A:        1
B:        2
C:        3
NIL
```

**Actual behavior**
It gives the following output, no matter with `rlwrap` or not:
```
A: 1
B:        2
C:        3
NIL
```
**Code at issue**

```lisp
(defun print-list(ls) (format t "~{~a:~10t~a~%~}" ls))
(print-list (list :a 1 :b 2 :c 3))
```

* Versions or commit hashes of Clasp on which you observed the problem (if you have built from up to date source, saying so is fine, you don't need to dig around for hashes)
commit 6b5234a398e2ae53302300deb9fd8c90ea9321dd built from source on Gentoo
* If relevant, your operating system and other aspects of your computing environment
Gentoo 17.1, llvm and clang using version 13.0.1
* If you believe the compiler may be at fault, relevant declarations, including optimize settings (use `(clasp-cltl2:declaration-information 'optimize)`)
* Any other context about the problem

yitzchak:
Fixes #1348

One remaining issue is what to do for set_column for CLOS streams since there isn't a corresponding generic in Gray streams. Do we add our own SETF or use generic_set_column which does nothing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 ## Fixes
 * Ensure that `print-unreadable-object` can accept output stream designators.
+* Set stream column to zero after printing the prompt in a REPL. Fixes [#1348][].
 
 # Version 2.1.0 (LLVM14) 2023-01-01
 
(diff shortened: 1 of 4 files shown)


Bike:
From a quick review of the gray streams protocol, it looks like it's pretty much set up so that the column is an intrinsic property of the stream, that's changed by writing characters and such. Making a `stream-line-column` writer would in contrast make the column more of a synthetic thing. Which of course it actually kind of is, but it still seems like a different view to me. Also, practically speaking nobody is implementing methods for such a writer as far as I know. So I'd lean towards not adding a writer. If an extension stream designer wants to handle the column they could do that themselves in `stream-read-line` and such, I think?

Bike (approved):

PR #1409, opened 2023-01-04
When poll/select indicate non-blocking then check for bytes

### LINKED ISSUE #1404: cl:listen not functioning on file streams
`touch` a file (here `foo.txt`) and try:

```lisp
(with-open-file (s "foo.txt") (listen s)) ; => T
```

This should return `nil` as there is no more content in the file.

yitzchak:
Fixes #1404

### DIFF
--- a/src/core/lispStream.cc
+++ b/src/core/lispStream.cc
@@ -380,8 +380,8 @@ void StreamCursor::backup(T_sp strm, claspCharacter c) {
 const FileOps &duplicate_dispatch_table(const FileOps &ops);
 const FileOps &stream_dispatch_table(T_sp strm);
 
-static int flisten(T_sp, FILE *);
-static int file_listen(T_sp, int);
+static int file_listen(T_sp, FILE *);
+static int fd_listen(T_sp, int);
 
 //    static T_sp alloc_stream();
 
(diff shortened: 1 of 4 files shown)


src/core/lispStream.cc
```diff
@@ -5152,11 +5100,94 @@ static int file_listen(T_sp stream, int fileno) {
...
+     3 and 4. */
+#elif defined(HAVE_SELECT)
+  fd_set fds;
+  struct timeval tv = {0, 0};
+  /*
+   * Note that the following code is fragile. If the file is closed (/dev/null)
+   * then select() may return 1 (at least on OS X), so that we return a flag
+   * saying characters are available but will find none to read. See also the
+   * code in cl_clear_input().
+   */
```
Bike, line 5134: I don't understand this comment. It makes it sound like the code is broken on `/dev/null`. But it looks like if `select` returns 1 we continue on to `ioctl` or `read` anyway, rather than returning a flag saying characters are available.

yitzchak: To be honest, I don't either. I didn't add it. It was already in Clasp. I think it might be obsolete.

yitzchak: I just removed the comment.

Bike (approved):

PR #1422, opened 2023-02-04
Update to LLVM 15

yitzchak:


include/clasp/core/core.h
```diff
@@ -295,6 +295,9 @@ const handleType UniqueIdHandle = 1;
 #define UndefinedUnsignedInt UINT_MAX
 #define UNDEF_UINT UndefinedUnsignedInt
 
+#define __EX(var) #var
+#define CXX_MACRO_STRING(var) __EX(var)
+
```
Bike, line 300: Why not just `#define CXX_MACRO_STRING(var) #var`? Actually, why this macro at all? An explanatory comment would be nice

Bike (approved):
Assuming we want to commit to LLVM15 now, looks fine to me.

include/clasp/core/core.h
yitzchak: You need double expansion to make CXX_MACRO_STRING(LLVM_VERSION) actually turn into "15.0"

PR #1426, opened 2023-02-09
Add some support for implementations other than SBCL in bootstrapping

yitzchak:
- [ ] abcl - Gray streams appear broken
- [x] ccl - `tail --lines=+2 koga | ccl --no-init --batch`
- [ ] clasp - This will take quite a bit more work. I'll have to add a switch for registering immutable systems and make ASDF ignore the host implementation's SYS logical paths. I'm gonna punt.
- [ ] clisp - Its doable, but there are a lot of changes. Some stuff in UIOP doesn't seem to work.
- [x] ecl - `ecl --norc --shell koga`
- [x] sbcl - `sbcl --script koga`


koga
```diff
@@ -122,7 +123,8 @@
   ;; Do the absolute minimum to inform ASDF about the location of systems
   ;; in order to find the clasp root and the desired build directory.
   (asdf:initialize-source-registry
-    `(:source-registry (:tree ,(uiop:getcwd))
+    `(:source-registry (:also-exclude "asdf")
```
Bike, line 126: Maybe should mention why asdf is excluded in a comment (i guess to avoid reloading it from an implementation asdf?)

src/koga/configure.lisp
```diff
@@ -513,10 +513,21 @@
...
         :documentation "The git binary to use.")
+   (lisp :accessor lisp
+         :initarg :lisp
+         :initform #+abcl #P"abcl"
+                   #+ccl #P"ccl"
+                   #+clasp #P"clasp"
+                   #+clisp #P"clisp"
+                   #+ecl #P"ecl"
+                   #+sbcl #P"sbcl"
+                   #-(or abcl ccl clasp clisp ecl sbcl) nil
```
Bike, line 524: if we load on some alien lisp, will NIL result in anything sensible later on? If I'm reading lisp-command correctly, it might end up trying to run a program called "NIL". if the results are weird, we should instead make the initform an error, or otherwise make a sensible error pop up.

src/koga/ninja.lisp
```diff
@@ -4,6 +4,20 @@
   (declare (ignore configuration name))
   (ninja:make-line-wrapping-stream (ninja:make-timestamp-preserving-stream path)))
 
+(defun lisp-command (script &optional arguments)
+  (concatenate 'string
+               "$lisp "
+               #+abcl "--noinform --batch --noinit "
```
Bike, line 10: If it is possible for the `:lisp` config argument to not be the implementation koga is running lisp-command on, the option flags ought to be based on the former rather than the latter. And if it is not possible there probably shouldn't be a `:lisp` config option.

Bike (changes requested):
Seems fine - as always I am nitpicking

koga
yitzchak: Yeah, it is excluded so the host uses their own. Mostly because ECL doesn't like the latest ASDF.

src/koga/configure.lisp
yitzchak: Thats a good point

src/koga/ninja.lisp
yitzchak: Right now it is just there in case the binary is not named something obvious. CCL does this sometimes with `ccl64`.

yitzchak:
No worries, nitpicking makes good code.

PR #1437, opened 2023-03-21
Callback

Bike:
Eliminates `defcallback` special operator, and provides a macroexpansion for `foreign-call-pointer`. This is enough to let CFFI work from the VM.


Bike:
test failure is that transient the.8 thing, which i guess i should file a bug for, but which is unrelated to these changes

PR #1451, opened 2023-04-13
Bytecode debug

Bike:
Adds some debugging capabilities for bytecode. Specifically, `disassemble` is slightly nicer, and more importantly, there is a "line tables" system that records where to find local variables. Precise source locations are not in yet, though.


Bike:
Juuuuust remembered I should update disltv to handle the attribute

PR #1457, opened 2023-05-20
Signal error for :ARGUMENTS in DEFINE-METHOD-COMBINATION at eval.

yitzchak:
Warn at expansion


Bike (approved):
We could possibly be more fancy and signal a program error with source location and form, but the warning will already have a source location, so I don't think it's a big deal, especially as a temporary measure.

PR #1458, opened 2023-05-22
ANSI test changes

yitzchak:
Use upstream versus our fork. The main change is that upstream now supports an expected failures file. I've changed the subset command since upstream doesn't have that. I've rolled it into koga versus managing it in our ansi-test fork.

To run all the tests `ninja -C build ansi-test`.  To run a subset `ANSI_TEST_SUITE=printer ninja -C build ansi-test`. Expected errors are stored in `tools-for-build/ansi-test-expected-failures.sexp`


Bike (approved):

Bike:
out of curiosity, does upstream ansi-test support running only subsets of the tests, like you can do with kpoeck's fork?

PR #1464, opened 2023-06-06
Bytecode source info

Bike:
Makes the bytecode file compiler record detailed source info within functions. This allows debugger frames to indicate precisely (or at least more precisely) which form was being evaluated when an error occurred.


Bike:
Superseded by #1465

PR #1468, opened 2023-06-16
Bytecode diagnostics

Bike:
Make the bytecode compiler signal warnings with accurate source locations on unknown functions and variables, and give compiler macro expansion warnings accurate source locations as well.


Bike:
Test failures are due to ASDF's undeferred-warnings test expecting Clasp with bytecode to not signal warnings. So, the failures aren't real. Merging.

PR #1469, opened 2023-07-02
Speed up filling vectors

drmeister:
Speed up filling int and float vectors


Bike:
Test failures aside - why does the from_object code for a float array need to worry about fixnums?

PR #1472, opened 2023-07-21
Bytecode to bir

Bike:
Defines a system for compiling bytecode directly into BIR (and thereafter into native code). Hooks it up to `cl:compile`, so that running `cl:compile` on a bytecoded function will compile it to native. Note that bytecoded functions are still considered compiled (i.e. are `compiled-function-p`). I think this behavior is kosher. There is no automatic compilation yet.

This is **experimental**. It is **not ready for prime time** and some of my design is pretty sloppy. But it's stable enough that I want to try using it as part of normal workflows so we can see what breaks.


Bike:
@drmeister has asked me not to merge this until we sort out what's going wrong with snapshots - they are not currently working on this branch, but not working on main either.

PR #1497, opened 2023-09-22
Add LLVM 17 support

yitzchak:
Uses new pass manager in `llvm-sys:optimize-module` and moves usage of legacy pass manager (still need for codegen) into `llvm-sys:emit-module`.


src/lisp/kernel/cmp/compile-file.lisp
```diff
@@ -144,27 +144,20 @@ and the pathname of the source file - this will also be used as the module initi
...
+                                                                  (llvm-sys:get-triple triple)
+                                                                  ""
+                                                                  ""
+                                                                  target-options
+                                                                  reloc-model
+                                                                  (code-model :jit nil :target-faso-file target-faso-file)
+                                                                  'llvm-sys:code-gen-opt-default
+                                                                  nil)
+                                  output-stream
+                                  nil #|<-dwo-stream|#
```
Bike, line 159: what is dwo stream? Doesn't seem to be documented in LLVM either from a quick search, but might be good to mention... sssssomewhere.

src/llvmo/llvmoExpose.cc
```diff
@@ -642,23 +632,38 @@ CL_DEFMETHOD core::T_sp TargetMachine_O::addPassesToEmitFileAndRunPassManager(Pa
...
     SIMPLE_ERROR("Illegal file type {} for addPassesToEmitFileAndRunPassManager", _rep_(dwo_stream));
   }
-  if (this->wrappedPtr()->addPassesToEmitFile(*passManager->wrappedPtr(), *ostream.get_stream(), dwo_ostream.get_stream(), FileType, true, nullptr)) {
+
+  llvm::legacy::PassManager PM;
+
+  llvm::LLVMTargetMachine* LTM = dynamic_cast<llvm::LLVMTargetMachine*>(TM);
+  if (LTM) {
+    llvm::TargetPassConfig* TPC = LTM->createPassConfig(PM);
+    TPC->setEnableTailMerge(false);
```
Bike, line 641: might be good to mention why we're doing this - iirc drmeister disabled tail merges so that debug info is more accurate. otherwise it just seems kind of random to do it here

Bike (approved):
Assuming this builds, and I'm sure it does, seems fine. You had me at +236 -1,401 honestly

src/llvmo/llvmoExpose.cc
yitzchak: Add comment for that.

src/lisp/kernel/cmp/compile-file.lisp
yitzchak: Apparently it is a Dwarf Object stream. Improved the comment.

PR #1503, opened 2023-09-30
Improve printing of quasiquote

yitzchak:


src/core/write_list.cc
```diff
@@ -44,102 +44,106 @@ THE SOFTWARE.
...
+
+  if (op == cl::_sym_function) {
+    if (tail)
+      return false;
+    clasp_write_string("#'", stream);
+    write_object(oCar(cdr()), stream);
+    return true;
+  }
+
+  T_sp quasiquote = _sym_STARquasiquoteSTAR->symbolValue();
```
Bike, line 69: Might be worth explaining what `*quasiquote*` is for, somewhere, briefly. If I've divined it correctly it's not that complicated - a plist from streams to a "are we currently writing a quasiquote form" marker - but it's nontrivial. Also, does it do okay with nested quasiquotation? As written it looks like printing `(core:unquote foo)` outside of quasiquotation will just print that instead of `,foo`, but how about ``(`(foo `(bar ,,,baz)))`` - will that print with an extra comma even though it has one too many unquotes, instead of as ``(`(foo `(bar (core:unquote ,,baz)))``? (Is this too pedantic to care about?)

src/lisp/kernel/lsp/pprint.lisp
```diff
@@ -705,7 +710,14 @@
 ;;;; Utilities.
 
 (defun pprint-pop-helper (object count stream &aux code)
-  (cond ((not (listp object))
+  (cond ((or (not (listp object))
+             (and (plusp count)
+                  (consp (cdr object))
+                  (null (cddr object))
+                  (or (eq (car object) 'core:quasiquote)
+                      (and (getf core:*quasiquote* (target-stream stream))
```
Bike, line 718: this getf is showing up enough that it might be worthwhile defining a `core:` function for it - `printing-quasiquote-p` or something - to hide the details of what's in `core:*quasiquote*` a bit. Though format-pprint up there would still have to bind it.

Bike (changes requested):
Other than the stuff I mentioned I think this looks good.

src/core/write_list.cc
yitzchak: Correct, it will handling nesting like that. I added a few regression tests that show a few levels of that.

yitzchak: Specifically, because getf returns the first key/value binding as `(list* stream nil core:*quasiquote*)` will ignore the trailing nestings.

src/lisp/kernel/lsp/pprint.lisp
yitzchak: I was going to try do a similar treatment for circle detection. Basically, make the 2-phase circle detection more robust in case of nesting. That is more complex and I need to think about it a bit more. So maybe wait on generalizing or abstracting this idea until then?

Bike: Works for me.

yitzchak: Cool. Thanks.

PR #1507, opened 2023-11-04
Hybrid kernel

yitzchak:


src/core/load.cc
```diff
@@ -178,29 +178,40 @@ CL_DEFUN T_sp core__load_no_package_set(T_sp lsource, T_sp verbose, T_sp print,
...
+          if (cl__equalp(key, pathname->_Type)) {
+            function = oCdar(_hooks);
+            break;
+          }
+        } else {
+          if (magic.nilp()) {
+            T_sp stream = cl__open(pathname, kw::_sym_input, ext::_sym_byte8, nil<T_O>(), false, nil<T_O>(), false, external_format,
+                                   nil<T_O>());
+            uint8_t bytes[4];
+            clasp_read_byte8(stream, bytes, 4);
```
Bike, line 194: Will this signal an error if the file is less than 4 bytes long? I'm not sure it does. Obviously that's a pretty obscure case but I can imagine accidentally trying to load an empty file.

Bike (changes requested):

src/core/load.cc
yitzchak: No, it will just return what it can read which won't match the 4 byte magic numbers. I tested this with a "touch" to make an empty file.

Bike: Works

Bike (approved):

PR #1508, opened 2023-11-06
Cleanup build

yitzchak:
- Remove `:object`, `:ll`, `:bc`, and `:fasl` build modes
- Replace `core:*clasp-build-mode*` and `cmp:*default-object-type*` with `cmp:*default-output-type*`
- Remove .fasp, .faspbc and .faspll file extension
- Replace CLASP_BUILD_MODE define with DEFAULT_OUTPUT_TYPE_BYTECODE, DEFAULT_OUTPUT_TYPE_FASO, etc
- Disable intrinsics-cxx.bc and other bitcode files that are unused in build
- Use single expected failures file for ansi-test
- Remove functions that aren't being used anymore

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,9 @@
 * New build mode `:bytecode-faso` which builds the kernel as native
   code (FASO) while the bytecode compiler is active.
 
+## Removed
+* Obsolete `:object`, `:ll`, `:bc`, and `:fasl` build modes.
+
 # Version 2.4.0 (LLVM15-17) 2023-10-01
 
 ## Added


src/lisp/regression-tests/system-construction.lisp
```diff
@@ -21,13 +21,13 @@
     (cmp::cfp-output-extension :fasl)
     (pathname-type (compile-file-pathname "test.lisp")))))
 
-(test-true compile-file-pathname-3a
+#+(or)(test-true compile-file-pathname-3a
```
Bike, line 24: if we no longer support bitcode output we should just axe this test code.

src/lisp/regression-tests/system-construction.lisp
```diff
@@ -49,7 +49,7 @@
 ;;; crosscompiling sbcl
 (test-true compile-file-serial-no-faso
  (let ((cmp::*compile-file-parallel* nil)
-       (core:*clasp-build-mode* :faso)
+       (cmp:*default-output-type* :faso)
```
Bike, line 52: it might be better to use the `:output-type` parameter to `compile-file` rather than the special variable.

src/llvmo/cscript.lisp
```diff
@@ -13,7 +13,7 @@
            #~"runtimeJit.cc"
            #~"clbindLlvmExpose.cc")
 
-(k:sources :bitcode
-           #~"intrinsics.cc"
-           #~"builtins.cc")
-
+; We probably don't need this anymore.
```
Bike, line 16: again, if we don't support bitcode builds, we should take the plunge and outright delete this.

Bike (approved):
Some minor stuff but I think this is basically great. So much old code gone

src/lisp/regression-tests/system-construction.lisp
yitzchak: Axed.

src/lisp/regression-tests/system-construction.lisp
yitzchak: Done

src/llvmo/cscript.lisp
yitzchak: Axed.

PR #1516, opened 2023-11-22
Add abort argument to close

yitzchak:


yitzchak:
I used a T_sp versus bool to avoid casting it back and forth, specifically for the concatenate streams, etc.

PR #1527, opened 2023-12-04
Misc fixes for Gray streams

yitzchak:
Some small tweaks to some default methods and making PATHNAME and TRUENAME generic functions. This is optional per the Gray stream proposal.

### DIFF
--- a/src/core/pathname.cc
+++ b/src/core/pathname.cc
@@ -888,7 +888,7 @@ struct PathnameRecursionGuard {
 };
 
 DOCGROUP(clasp);
-CL_DEFUN Pathname_sp cl__pathname(T_sp x) {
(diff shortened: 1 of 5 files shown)


src/core/pathname.cc
```diff
@@ -911,6 +911,12 @@ CL_DEFUN Pathname_sp cl__pathname(T_sp x) {
   return gc::As<Pathname_sp>(x);
 }
 
+CL_DOCSTRING("Returns the pathname denoted by pathspec.")
```
Bike, line 914: please put in a comment or something here and on cl__truename explaining the odd definition (that these will be made generic later)

Bike (changes requested):

PR #1531, opened 2023-12-17
Convert lisp stream dispatch to virtual methods

yitzchak:


src/core/lispStream.cc
```diff
@@ -1367,331 +992,110 @@ static int utf_8_encoder(T_sp stream, unsigned char* buffer, claspCharacter c) {
...
-  T_sp fn = eval::funcall(gray::_sym_stream_read_sequence, strm, data, clasp_make_fixnum(start), clasp_make_fixnum(end));
-  if (fn.fixnump()) {
-    return fn.unsafe_fixnum();
+T_sp FileStream_O::string_length(T_sp string) {
+  int l = 0;
+  if (cl__characterp(string)) {
+    l = compute_char_size(string.unsafe_character());
+  } else if (cl__stringp(string)) {
+    Fixnum iEnd;
+    String_sp sb = string.asOrNull<String_O>();
```
Bike, line 1019: Obviously this is just code you're copying and so fixing it is a bit out of scope, but just to mention (you probably already know):
1) asOrNull is to be avoided. we have `gc::As` etc
2) this conditional is pointless, since `cl__length` will return the fill pointer
3) ditto StringFillp and SetStringFillp. they should be on the chopping block

src/core/lispStream.cc
```diff
@@ -4762,19 +3339,16 @@ CL_DOCSTRING(R"dx(interactive_stream_p)dx");
...
   /* ANSI and Cltl2 specify that open-stream-p should work
            on closed streams, and that a stream is only closed
            when #'close has been applied on it */
-  if (Instance_sp instance = strm.asOrNull<Instance_O>()) {
-    return eval::funcall(gray::_sym_open_stream_p, instance);
-  }
-  unlikely_if(!AnsiStreamP(strm)) ERROR_WRONG_TYPE_ONLY_ARG(cl::_sym_open_stream_p, strm, cl::_sym_Stream_O);
-  return (StreamClosed(strm) ? nil<T_O>() : _lisp->_true());
+  AnsiStream_sp ansi_stream = strm.asOrNull<AnsiStream_O>();
+  return ansi_stream ? (ansi_stream->_Closed ? nil<T_O>() : _lisp->_true()) : eval::funcall(gray::_sym_open_stream_p, strm);
```
Bike, line 3351: maybe oughta have a reader rather than using a field directly

Bike (approved):
Good work.

PR #1540, opened 2023-12-25
Stream updates

yitzchak:
- Add primitive stream_file_descriptor. We inherited GRAY:STREAM-FILE-DESCRIPTOR from ECL. It was introduced only to port SERVE-EVENT. It is not in any other implementation and was never hooked up to any of the internal ansi-stream stuff. I've changed the signature because optional parameters on generic functions are not very useful especially they can't be specialized. The DIRECTION argument is now any direction which is valid with CL:OPEN.
- Add source input tracking
- Add a pseudo echo mode for two-way-stream. This updates the output cursor for values that are read from the input stream, assuming that the underlying streams are doing the actual echoing. I got the idea from CCL which has ECHOING-TWO-WAY-STREAM which does the same thing.
- Added `EXT:+PROCESS-TERMINAL-IO+` which is the actual two-way-stream behind `CL:*TERMINAL-IO*`, `CL:*DEBUG-IO*`, and `CL:*QUERY-IO*`. Turned those streams into synonym streams.


src/lisp/kernel/clos/streams.lisp
```diff
@@ -240,20 +245,23 @@ truename."))
...
 
-(defgeneric stream-file-descriptor (stream &optional direction)
+(defgeneric stream-file-descriptor (stream direction)
   (:documentation
    "Return the file-descriptor underlaying STREAM, or NIL if not
-   available. DIRECTION must be either :INPUT, or :OUTPUT and is
-   supposed to discriminate in case STREAM is a bidirectional
-   stream. DIRECTION is supposed to default to :INPUT.
+   available. DIRECTION must be either :INPUT, :OUTPUT, :IO or
+   :PROBE."))
```
Bike, line 252: Maybe restore the part of the comment explaining why there's a direction parameter?

src/lisp/kernel/clos/streams.lisp
```diff
@@ -812,42 +816,33 @@ truename."))
...
-(defmethod stream-file-descriptor ((stream file-stream) &optional (direction
-                                                                   :input))
-  (declare (ignore direction))
-  (ext:file-stream-file-descriptor stream))
+(defmethod stream-file-descriptor (stream direction)
+  (declare (ignore stream direction))
+  nil)
+
+(defmethod stream-file-descriptor ((stream ansi-stream) direction)
+  (%stream-file-descriptor direction))
```
Bike, line 824: Shouldn't this call get the stream as an argument?

src/lisp/kernel/clos/streams.lisp
```diff
@@ -812,42 +816,33 @@ truename."))
...
+
+
+;;; STREAM-INPUT-COLUMN
+
+(defmethod stream-input-column (stream)
+  (declare (ignore stream))
+  nil)
+
+(defmethod stream-input-column ((stream ansi-stream))
+  (%stream-input-column direction))
```
Bike, line 834: Should have the stream rather than the direction, surely. If the tests don't cover this maybe they ought to.

src/lisp/kernel/clos/streams.lisp
```diff
@@ -812,42 +816,33 @@ truename."))
...
+
+
+;;; STREAM-INPUT-LINE
+
+(defmethod stream-input-line (stream)
+  (declare (ignore stream))
+  nil)
+
+(defmethod stream-input-line ((stream ansi-stream))
+  (%stream-input-line direction))
```
Bike, line 844: should have the stream and not a direction.

src/core/lispStream.cc
```diff
@@ -566,16 +566,76 @@ int stream_set_column(T_sp stream, int column) {
...
+  if (stream.isA<AnsiStream_O>())
+    stream.as_unsafe<AnsiStream_O>()->restore_output_cursor();
+}
+
+// Stream input tracking functions
+
+// This function is exposed to CL because it is needed to implement
+// the GRAY:STREAM-INPUT-COLUMN for ansi-stream. The stream argument
+// is guaranteed to be an AnsiStream_sp so recursion is avoided.
+CL_LISPIFY_NAME("gray:%stream-input-column")
```
Bike, line 606: I'm mildly concerned about these functions being exposed in the GRAY package. I see that they are unexported later, but maybe it would be cleaner to move them to CORE, or for them to be defined in GRAY but not exported (I think there's a C macro to use instead of `SYMBOL_EXPORT_SC_` for that?)

Bike (changes requested):

src/lisp/kernel/clos/streams.lisp
yitzchak: You are correct.

src/lisp/kernel/clos/streams.lisp
yitzchak: More copy/paste errors.

src/core/lispStream.cc
yitzchak: I moved them from core because only gray is using them and they seem to be an internal design detail of the gray package. The scrapper and expose_function export all function automatically. I started down the patch of adding a "internal" option, but it looks like that will take more work and is kind of a tangential issue.

PR #1572, opened 2024-04-06
Simplify Integer_O::create and add const reference translators

paulapatience:


yitzchak:
@paulapatience  Ignore the variable length array warnings. These are LLVM18 warnings that I haven't fixed yet.

Bike (approved):

PR #1575, opened 2024-04-15
Fasl native

Bike:
This was supposed to have native FASL, but I get a pretty inscrutable failure in FASO build mode. So I'm going to merge this and then think about something else for a while.


Bike:
I decided not to keep the last defvar commit in faso-cleanup since it might be detrimental sometimes and the speedup is marginal.

PR #1582, opened 2024-05-12
Remove MPS from preciseprep builds

yitzchak:
Leave the defines in for reference during MMTK integration.

### DIFF
--- a/repos.sexp
+++ b/repos.sexp
@@ -204,10 +204,6 @@
   :repository "https://github.com/clasp-developers/asdf.git"
   :directory "src/lisp/modules/asdf/"
   :branch "add-clasp-bytecode-support")
- (:name :mps
-  :repository "https://github.com/Ravenbrook/mps.git"
-  :directory "src/mps/"
-  :commit "b8a05a3846430bc36c8200f24d248c8293801503")
  (:name :bdwgc
   :repository "https://github.com/ivmai/bdwgc.git"
   :directory "src/bdwgc/"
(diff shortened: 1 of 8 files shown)


repos.sexp
```diff
@@ -204,7 +204,7 @@
   :repository "https://github.com/clasp-developers/asdf.git"
   :directory "src/lisp/modules/asdf/"
   :branch "add-clasp-bytecode-support")
- (:name :mps
+ #+(or)(:name :mps
```
Bike, line 207: Can we not just remove this entirely?

src/gctools/cscript.lisp
```diff
@@ -10,7 +10,7 @@
            #~"exposeClasses3.cc"
            #~"boehmGarbageCollection.cc"
            #~"mmtkGarbageCollection.cc"
-           #~"mpsGarbageCollection.cc"
+           ;#~"mpsGarbageCollection.cc"
```
Bike, line 13: If we don't ever need it we should just delete this file. MMTk works differently from MPS so I don't think it will be much help as a reference, and even if we do need it, the file will always be there in the git history.

Bike (changes requested):
If we can totally remove all remnants of MPS we should. The GC code we have is messy and I don't think it will be helpful as a reference for MMTk or otherwise, and even if we do need it we can just check git.

yitzchak:
@Bike I removed the commented stuff in repos.sexp and the cscript files, at least.

PR #1588, opened 2024-05-31
Add Guix packages for Clasp and Cando

paulapatience:


guix/cando.scm
```diff
@@ -0,0 +1,101 @@
...
+              ;; <https://github.com/cando-developers/cando/commit/85ffab474faa342c38ced5b84553db035d23687b>.
+              (add-after 'unpack 'fix-build
+                (lambda _
+                  (substitute*
+                      '("extensions/cando/src/geom/matrix.cc"
+                        "extensions/cando/src/geom/ovector3.cc"
+                        "extensions/cando/src/geom/vector3.cc"
+                        "extensions/cando/src/kinematics/bondedJoint.cc")
+                    (("\\(isnan\\(") "(std::isnan(")
+                    (("\\|isnan\\(") "|std::isnan("))))))))
```
yitzchak, line 89: Do we still need these substitutions?

paulapatience: You're right, we don't. I hadn't realized that repos.sexp followed Cando's main branch.

yitzchak: Yes, When we do a release the repos.sexp in the source archive will have a specific commit for all dependencies.

PR #1599, opened 2024-06-21
Gentoo distro build script for 2.6.0

mgsouth:
New "ebuild" script used to create Gentoo Portage packages. Includes metadata & Manifest files.

Tested with both settng of the only flag (selects :faso or :bytecode-faso). Ran w/ regression tests, although they are not enabled in the distro package.

Users will need to download these three files to a local repository (instructions for createing local repos are on Gentoo wiki).

REPOPATH is path to that repository, such as "/var/db/repos/myrepo".

REPONAME is its label for Portage purposes.

As root:

```
mkdir -p REPOPATH/dev-lisp/clasp
(Download the three files in https://github.com/clasp-developers/clasp/gentoo/
 to REPOPATH/dev-lisp/clasp)
egencache --repo=REPONAME --update dev-lisp/clasp
egencache --repo=REPONAME --update-pkg-desc-index
egencache --repo=REPONAME --update-manifests --sign-manifests=n
emerge --ask -j 2 dev-lisp/clasp
```
Job limit can be  bumped up if user has the RAM. I saw about 1GB used per core.


yitzchak:
Does it need to be named with the version number in the files name? This creates a problem because we would have to rename them constantly. Also, we can't store a file with hashes for the tarball in the repo, since the tarball is generated from the repo. We should probably have a script that generates the manifest instead.

mgsouth:
It's worse than renaming them. You have to create a new one (minor edits), and the Manifest file must be recreated. Should probably keep the older ones for as long as those particular tarballs are available.

mgsouth:
Uh. Hm. chicken-and-egg here. Maybe the .ebuild, metadata, and Manifest should be external to the tarball.

mgsouth:
It's possible for the user to generate the Manifest themselves. In fact, my install steps might already be doing it. So: No manifest in tarball. If not already doing it, have step for user to create it--those steps go into the Wiki page (build/install).

Again, hopefully Gentoo will pull the ebuild in. Then it will be on "them" (who might even be me) to update as part of normal distro maintenance.

src/koga/archive.lisp
```diff
@@ -40,4 +40,70 @@
...
+                              ,(mk-os-path dest-dir "clasp-9999.ebuild"))
+                    ("MISC" ,(mk-os-path src-dir "metadata.xml")
+                            ,(mk-os-path dest-dir "metadata.xml"))
+                    ,@(when archive `(("DIST" nil ,(mk-os-path archive))))))
+           ;; --
+           (manifest-path (os->cl-path (mk-os-path dest-dir manifest-name)))
+           (manifest-temp-path (uiop:tmpize-pathname manifest-path)))
+      (message nil "Creating Gentoo/Portage distro files for ~a..." pkg-ver)
+      (uiop:ensure-all-directories-exist (list manifest-temp-path))
+      (uiop:with-output-file (mfst-dest manifest-temp-path :if-exists :supersede)
```
yitzchak, line 87: Please don't use the uiop:with-output-file/uiop:with-input-file functions. They are just syntactic sugar.

yitzchak:
Once you have fix the remaining minor issues please squash the commits.

src/koga/archive.lisp
```diff
@@ -40,4 +40,70 @@
...
+              for line = (read-line mfst-src nil nil)
+              while line
+              ;; Omit manifest entries we're about to create
+              unless (some (lambda (x) (search (os-path-fname (third x)) line)) flist)
+                do (write-line line mfst-dest))))
+        ;; Copy files, add them to manifest
+        (loop for (mfst-type s-os-path  d-os-path) in flist
+              when s-os-path
+                do (copy-f s-os-path d-os-path)
+              do (uiop:safe-format! mfst-dest "~a ~a ~d BLAKE2B ~a SHA512 ~a~%"
```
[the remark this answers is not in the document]
mgsouth: Done

src/koga/archive.lisp
mgsouth: Done

mgsouth:
Squashed.  Didn't know the requester could do that to an in-progress PR...

yitzchak:
Merged. Well done!

PR #1600, opened 2024-06-21
A manpage for Clasp.

mgsouth:
Because the Gentoo package just seemed naked without one. (It won't be in the 2.6.0 package, but will be in any following ones.)

### DIFF
--- a/docs/clasp.1
+++ b/docs/clasp.1
@@ -1,0 +1,153 @@
+.\"
+.\" man page introduction to Clasp
+.\"
+.\" Copyright (c) 2024, Clasp Developers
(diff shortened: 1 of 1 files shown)


yitzchak:
I think this is probably a good start. @bike do you have any comments?

Bike:
Seems like a good start, yeah. Maybe we should also mention that `--help` displays further command line options

mgsouth:
Done. Added --version and a couple of env. vars too.

yitzchak:
Let's get a squash here too.

yitzchak:
Thanks!

Issue #1596, opened 2024-06-17
Extended characters trip up packages


Bike:
Various parts of the package code assume that package names are base strings. Example problem:

```lisp
 (defpackage "ゆびわ")
; =>
#<PACKAGE \375\277\277\277\276\206s\375\277\277\277\276\217>
```

PR #1602, opened 2024-06-22
Package reform

Bike:
Cleans some stuff up and, importantly, allows package names that aren't just base strings (#1596).

Unfortunately does **not** fix the way package objects are printed - which is relatively minor, but annoying. The problem is it goes through __repr__ and that returns a std::string.


Bike:
Fixed it up so package names print right now, so I think it fixes #1596.

PR #1628, opened 2024-09-05
Fix MAP.ERROR.11

yitzchak:
Do a type check for result types like `(or (vector t 5) (vector t 10))` in MAP. Tries to avoid doing the type check for simple type specifiers like VECTOR or specialized array classes.

CMUCL has the same bug we do in MAP and this is basically how SBCL fixed it. Maybe not ideal, but I think it is probably the easiest fix without changing the plumbing in MAKE-SEQUENCE significantly.


Bike (approved):
right, this... i thought about doing this a while ago but didn't want to slow things down until the compiler was better, but it's better for it to be correct first.
I'll probably need to add type derivers for make-sequence and what it expands into, later.

PR #1631, opened 2024-09-16
Add long-float

yitzchak:


include/clasp/core/bignum.h
```diff
@@ -33,6 +33,19 @@ THE SOFTWARE.
 #include <concepts> // integral
 
 namespace core {
+
+template <class T>
+concept unsigned_limb = std::is_integral<T>::value && std::is_unsigned<T>::value && sizeof(T) <= sizeof(mp_limb_t);
```
Bike, line 38: this all is pretty neat.

include/clasp/core/bignum.h
```diff
@@ -69,35 +82,55 @@ class Bignum_O : public Integer_O {
...
-  static Bignum_sp create(int64_t v) { return create_from_limbs((v < 0) ? -1 : 1, std::abs(v), true); }
-#endif
-  static Bignum_sp create(uint64_t v) { return create_from_limbs(1, v, true); }
-  static Bignum_sp create(__uint128_t v) {
-    Bignum_sp b = create_from_limbs(2);
-    b->_limbs[0] = static_cast<mp_limb_t>(v);
-    b->_limbs[1] = static_cast<mp_limb_t>(v >> 64);
+
+  template <unsigned_limbs T> static Bignum_sp create(T v) {
+    constexpr size_t limb_width = 8 * sizeof(mp_limb_t);
```
Bike, line 99: obviously it's not going to matter conformancewise, but i think we should use CHAR_BIT instead of 8 just to get the point across.

include/clasp/core/object.h
```diff
@@ -511,15 +511,13 @@ inline CL_DEFUN bool cl__eql(T_sp x, T_sp y) {
...
   } else if (x.single_floatp()) {
-    if (y.single_floatp()) {
-      return gc::tagged_single_float_masked(x.raw_()) == gc::tagged_single_float_masked(y.raw_());
-    }
-    return false;
+    if (!y.single_floatp())
+      return false;
+    single_float_t xf = x.unsafe_single_float();
+    single_float_t yf = y.unsafe_single_float();
+    return xf == yf && std::signbit(xf) == std::signbit(yf);
```
Bike, line 518: probably worth mentioning the point of this (signed zero comparisons) in a comment, since it's a bit inscrutable

src/lisp/kernel/cleavir/type.lisp
```diff
@@ -1043,6 +1062,8 @@
             (ctype:range 'single-float 0f0 '* sys))
            ((subtypep max (ctype:range 'double-float 0d0 '* sys))
             (ctype:range 'double-float 0d0 '* sys))
+           ((subtypep max (ctype:range 'long-float 0d0 '* sys))
```
Bike, line 1065: should be `0l0`?

include/clasp/core/array.h
```diff
@@ -224,7 +226,7 @@ class Array_O : public General_O {
   /*! length() doesn't dispatch - it reaches into the subclass
       through the _Length[0] array to read the first size_t element
       which is the Length/FillPointer for vectors and a Dummy value for arrays */
-  size_t length() const { return this->_Length[0]; };
+  virtual size_t length() const { return this->_Length[0]; };
```
Bike, line 229: I don't understand this change. Why does this have to work differently for short/long arrays? Is there some way for _Length[0] and the gcarray length to get out of sync? If I remember correctly, the idea here is that for a 1d simple array _Length[0] reaches into the underlying gcarray anyway.

If this change is necessary, the above comment saying "length() doesn't dispatch" needs to be changed, since if length is virtual the comment is false.

Bike (commented):
mostly good. You sure worked a lot of semi related stuff into this PR, but it's all cleanup that needed doing anyway

include/clasp/core/bignum.h
yitzchak: Good point

src/lisp/kernel/cleavir/type.lisp
yitzchak: Yes.

include/clasp/core/array.h
yitzchak: I didn't want make the change, but I'm not sure what to do. I think that because x86 stores binary80 with 6 bytes of padding that affects the alignment of the structure so the non-dispatching hack doesn't work anymore. I'm not sure why we have such a hack anyways?

Bike: well the hack is basically so that getting the length of an array is a simple memory access regardless of type. If the length is no longer in that position we might also have to rejigger the compiler; at least for other simple arrays it assumes it can just grab the length without calling a function.

yitzchak: Fixed with packed attribute.

Bike (approved):

PR #1640, opened 2024-10-19
koga enhancements: defines, *.cxx and *.hpp as sources.

dg1sbg:
This PR brings the following enhancements to koga:

1. Add source files with endings .cxx and .hpp as valid C++ source and header files.
2. Allow for #define statements being added in cscript files: `(koga:defines "XYZ=ABC") ` adds `-DXYZ=ABC` to cflags.

The PR also includes a minor correction for a debug message saying that the locaton of an extension is taken from its repo definiton.


yitzchak:
Thanks!

PR #1642, opened 2024-10-19
koga enhancement: Allow #'library function to have a path spec.

dg1sbg:
Expanded koga'a #'library function (actually, #'configure-library) to support the following way of specifying a libary:

`(k:library "nddscpp2d" :path #p"/Applications/rti_connext_dds-7.3.0/lib/arm64Darwin20clang12.0/" :kind :dynamic)`

The path will be added as -L_path_  to the linker flags.

### DIFF
--- a/src/koga/configure.lisp
+++ b/src/koga/configure.lisp
@@ -908,33 +908,64 @@ the function to the overall configuration."
 -d suffix if debugging is enabled."
   (format nil "~a~:[~;-d~]" (variant-name variant) (variant-debug variant)))
 
+(defun library-info (library path kind)
+  (flet ((lib-ext ()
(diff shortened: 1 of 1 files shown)


yitzchak:
Clasp currently doesn't need non-pkgconfig support. Some user's that have built extensions that needed to use libraries that don't have pkgconfig support. In that case directly adding the appropriate ldflags is probably the easiest route. If can finish up dynamic extension support then this will be moot since extensions won't need to be built directly into Clasp, i.e. extensions will just link against libclasp.so. 

In other words, I'd prefer to just expose the needed ldflags functions if we must in the short term.

dg1sbg:
As @yitzchak pointed out, there will be another approach to handle pkconfig-less external libraries.

PR #1654, opened 2024-12-09
GC cleanup

Bike:
Various miscellaneous cleanup of the garbage collection/memory infrastructure as I prep to use Whippet. Shouldn't have any real externally visible changes except that ROOM doesn't ~~test dladdr or~~ output some information about classes. More to come, but I want to merge this before embarking on what might end up being a pretty big rewrite of hash tables.


Bike:
Various miscellaneous cleanup of the garbage collection/memory infrastructure as I prep to use Whippet. Shouldn't have any real externally visible changes except that ROOM doesn't output some information about classes. More to come, but I want to merge this before embarking on what might end up being a pretty big rewrite of hash tables.

PR #1681, opened 2025-04-17
Fix clang errors on mac, hopefully

Bike:
Errors mostly relating to `std::ranges`, seen here https://github.com/clasp-developers/clasp/actions/runs/14520823855


Bike:
The standard library on the mac runners just seems to be messed up. I don't see how it can not know about `std::ranges` unless it's out of date or something.

PR #1765, opened 2026-04-22
Interpreter work

Bike:
- `bytecode_vm` now maintains its own stack of dynenvs instead of calling itself recursively
- `bytecode_vm` uses computed goto rather than a big switch
- fixes up trampolines (i.e. native stubs for bytecode functions, so they show up in DWARF metadata for profiling etc.) to use copied object code rather than recompiling llvm-ir anew for each trampoline

very much a draft - i'm just filing it as a PR to review it


include/clasp/gctools/threadlocal.h
```diff
@@ -78,10 +78,39 @@ extern int global_debug_virtual_machine;
...
 #endif
 
+// ---------- Dynamic-environment records for the bytecode interpreter ----------
+// The bytecode VM establishes dynamic environments (tagbody, catch,
+// special-bind, progv, unwind-protect) by pushing records onto a side stack
+// instead of recursing into bytecode_vm. The entering opcodes push; the
+// matching exit opcodes pop; an outer try/catch(Unwind&) in bytecode_vm walks
+// the stack to run cleanups / resume at a saved pc on non-local exits.
+//
+// Currently only the type and the stack exist — no opcodes are migrated yet.
```
Bike, line 88: This entire parallel structure of dynenvs is pointless duplication. Even if we do want to keep the parallel stack of dynenvs, which I don't think we do, the kind, frame, slots, and all are apparent from the existing dynenv classes (TagbodyDynEnv_O etc.). A separate enum for what are types, padding bytes (??), slots that are punned, this is all really unacceptable. Only the bytecode specific bits like the sp mark and target pc should really need recording anyway, and we could probably just add them to the dynenv classes or make new ones.

Adding to dynenv classes might also remove the need for VMFrameDynEnv which would be good.

include/clasp/gctools/threadlocal.h
```diff
@@ -101,6 +130,13 @@ struct VirtualMachine {
   core::T_O** _literals;
   unsigned char* _pc;
 
+  // Dynamic-environment record stack. Root-allocated so GC scans the
+  // T_O*/T_O** slots conservatively. _dynRecordTop points one past the last
```
Bike, line 134: All of the slots in the parallel dynenv structures already exist in the dynenv classes, which are already reachable from the control stack or from the thread local state, so GC should not be a concern here

include/clasp/llvmo/code.h
```diff
@@ -84,6 +84,13 @@ class ObjectFile_O : public LibraryBase_O {
   size_t _Size;
   size_t _ObjectId;
   JITDylib_sp _TheJITDylib;
+  // If true, this ObjectFile is transient arena-init scaffolding (shared
+  // trampoline / stub template) that must not be serialized into snapshots.
+  // The ObjectFile is still registered in _AllObjectFiles normally — LLVM's
+  // link layer plugin looks it up by name during materialization, so it must
```
Bike, line 90: _Our_ link layer plugin looks it up by name during materialization, not LLVM. And it does so for reasons trampolines don't need - so that we can look up DWARF from an instruction pointer. Since trampolines are for bytecode functions, we can use the bytecode debug info mechanisms instead of using DWARF at all. So I don't think trampolines need to go in _AllObjectFiles.

EDIT: Okay, so actually these changes use the trampoline to get closure etc, for some reason, so we do need DWARF. However to figure out if a PC is in an arena we can use `arena_lookup_by_pc` so we're still not going through object files.

include/clasp/llvmo/trampolineWork.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+/*
+    File: trampoline.h
```
Bike, line 4: filename is incorrect

include/clasp/core/lisp.h
```diff
@@ -279,6 +279,10 @@ class Lisp {
     std::atomic<T_sp> _AllObjectFiles;
     std::atomic<T_sp> _AllCodeBlocks;
     std::atomic<T_sp> _AllBytecodeModules;
+    // Every GFBytecodeSimpleFun ever made (atomic-pushed list of cons cells).
```
Bike, line 282: "list of cons cells" indicates that this is a list whose elements are conses, but actually they are GFBytecodeSimpleFuns as suggested by the name.

include/clasp/core/lisp.h
```diff
@@ -279,6 +279,10 @@ class Lisp {
...
+    // Walked by arena_post_load_regenerate_trampolines after a snapshot load
+    // so the dispatch trampoline for each generic function gets re-attached.
+    std::atomic<T_sp> _AllGFBytecodeFuns;
```
Bike, line 285: maybe should be a List_sp

src/core/trampoline/trampoline.cc
```diff
@@ -21,7 +21,14 @@ void *CLASP_LITERALS(trampoline)[0];
 // Use asm and nodebug to add declaration for the LLVM intrinsic.
 __attribute__((nodebug)) void LLVM_EXPERIMENTAL_STACKMAP(uint64_t type, uint32_t dummy, ...) asm("llvm.experimental.stackmap");
 
-return_type bytecode_call(uint64_t pc, void *closure, uint64_t nargs, void **args);
+// Indirect through a global function pointer rather than calling bytecode_call
+// directly. Each compiled trampoline's call topology is now identical
```
Bike, line 25: what is "call topology"?

src/core/bytecode.cc
```diff
@@ -215,6 +215,70 @@ static Function_sp fdesignator_in_env(T_sp desig, T_sp env) {
...
+bytecode_vm(VirtualMachine& vm, T_O** literals, T_O** closed, Closure_O* closure,
+            core::T_O** fp, core::T_O** sp, size_t lcc_nargs, core::T_O** lcc_args);
+
+// Pop the topmost VMDynRecord and undo its effect. Used by the `unbind`
+// opcode, the `_return` defensive drain, and the outer catch handler — all
+// need to reverse one record worth of dynamic state.
+static inline void vm_pop_dyn_record(VirtualMachine& vm) {
+  VMDynRecord* r = --vm._dynRecordTop;
+  switch (r->kind) {
+  case VMDynKind::SpecialBind: {
```
Bike, line 229: This duplicates `DynEnv_O::proceed` for no reason.

src/core/bytecode.cc
```diff
@@ -242,6 +306,113 @@ bytecode_vm(VirtualMachine& vm, T_O** literals, T_O** closed, Closure_O* closure
...
+  VMDynRecord* const dyn_entry_mark = vm._dynRecordTop;
+
+  while (true) {
+    try {
+
+// ---------- Computed-goto dispatch ----------
+// GCC and Clang support the labels-as-values extension (&&label).
+// Each opcode ends with VM_NEXT which jumps directly to the next
+// opcode, avoiding the single indirect-branch bottleneck of a switch.
+// Disabled when DEBUG_VIRTUAL_MACHINE is on so the switch is preserved
```
Bike, line 324: I don't see how computed goto would prevent debugging or the recording mechanism.

src/core/bytecode.cc
```diff
@@ -242,6 +306,113 @@ bytecode_vm(VirtualMachine& vm, T_O** literals, T_O** closed, Closure_O* closure
...
+#if defined(__GNUC__) && !defined(DEBUG_VIRTUAL_MACHINE)
+#define USE_COMPUTED_GOTO 1
+#endif
+
+#ifdef USE_COMPUTED_GOTO
+  #define VM_CASE(name)  vm_op_##name
+  #define VM_NEXT        goto *dispatch_table[*pc]
+  static void* dispatch_table[256];
+  static bool dispatch_table_init = false;
+  if (!dispatch_table_init) [[unlikely]] {
```
Bike, line 335: Lazily initializing this is bad. This is another branch in a function that's used for every single (bytecode) Lisp function call, and the dispatch table is static so it shouldn't need runtime initialization at all.. Writing out a brace initialization would look very redundant with the repeated `&&vm_op_error`, but that can be hidden with macrology.

src/core/bytecode.cc
```diff
@@ -242,6 +306,113 @@ bytecode_vm(VirtualMachine& vm, T_O** literals, T_O** closed, Closure_O* closure
...
+    for (int i = 0; i < 256; ++i) dispatch_table[i] = &&vm_op_error;
+    dispatch_table[(uint8_t)vm_code::ref]                      = &&vm_op_ref;
```
Bike, line 337: `vm_code`s are already `uint8_t`s, so why the casts? And these are very redundant & could again be clearer with macros.

src/core/bytecode.cc
```diff
@@ -242,6 +306,113 @@ bytecode_vm(VirtualMachine& vm, T_O** literals, T_O** closed, Closure_O* closure
...
+    dispatch_table_init = true;
+  }
+  VM_NEXT; // initial dispatch
+
+  vm_op_error: {
+    SimpleFun_sp ep = closure->entryPoint();
+    BytecodeModule_sp bcm = gc::As<BytecodeSimpleFun_sp>(ep)->code();
+    unsigned char* codeStart = (unsigned char*)bcm->bytecode()->rowMajorAddressOfElement_(0);
+    unsigned char* codeEnd = codeStart + bcm->bytecode()->arrayTotalSize();
+    SIMPLE_ERROR("Unknown opcode {} pc: {}  module: {} - {}", *pc, (void*)pc, (void*)codeStart, (void*)codeEnd);
```
Bike, line 411: error message should probably just include the function, since the pc and module addresses are not going to be understood easily

src/core/backtrace.cc
```diff
@@ -433,44 +434,86 @@ static DebuggerFrame_sp make_lisp_frame(size_t frameIndex, void* absolute_ip, co
                                nil<T_O>(), INTERN_(kw, lisp), XEPp);
 }
 
-static DebuggerFrame_sp make_bytecode_frame_from_function(BytecodeSimpleFun_sp fun, void* bpc, T_O** bfp) {
+// Recover (closure, nargs, args) from the parent trampoline frame, if the
+// bytecode_call C++ frame at `bytecode_call_fbp` was reached via an arena
+// trampoline. The trampoline saves these at fixed negative offsets from its
```
Bike, line 439: this hardcodes the offsets that we expect LLVM to put various values in generated code. Not acceptable

Also, this will break backtraces of bytecode functions if unique trampolines are not in place (e.g. fallback on macs)

src/core/funcallableInstance.cc
```diff
@@ -522,6 +552,24 @@ std::string GFBytecodeSimpleFun_O::__repr__() const {
...
+  // (arena_post_load_regenerate_trampolines) walks _AllGFBytecodeFuns and
+  // re-installs a fresh per-GF arena trampoline.
+  if (snapshotSaveLoad::operation(fixup) == snapshotSaveLoad::SaveOp
+      && llvmo::arena_owns_pc((uintptr_t)this->_EntryPoints._EntryPoints[0].load(std::memory_order_relaxed))) {
+    this->_EntryPoints._EntryPoints[0].store(
+        (ClaspXepAnonymousFunction)g_gf_dispatch_entry_point_n, std::memory_order_relaxed);
+  }
+  // _Trampoline isn't actually invoked at runtime for GFBytecodeSimpleFun
+  // (dispatch goes through entry_point_n via _EntryPoints[0]), but keep the
+  // substitution as defense-in-depth in case anything ever reads the field.
```
Bike, line 568: It appears that nothing actually reads `_Trampoline`, at least before these changes, so if nothing in these changes is using the field either it should just be deleted.

src/core/function.cc
```diff
@@ -153,6 +159,24 @@ void BytecodeSimpleFun_O::fixupInternalsForSnapshotSaveLoad(snapshotSaveLoad::Fi
...
+  // Substitute the default bytecode_call (a libclasp symbol the encoder can
+  // handle) before the standard fixup runs; the post-load pass in llvmo
+  // (arena_post_load_regenerate_trampolines) re-attaches a fresh arena
+  // trampoline if the user is running with the arena backend.
+  if (snapshotSaveLoad::operation(fixup) == snapshotSaveLoad::SaveOp
+      && llvmo::arena_owns_pc((uintptr_t)this->_Trampoline)) {
+    static std::atomic<size_t> s_subbed{0};
+    size_t n = s_subbed.fetch_add(1) + 1;
+    if (n <= 3 || (n % 10000) == 0) {
+      fprintf(stderr,
```
Bike, line 173: unconditional debug output

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,487 @@
+/*
+ * sampling_profiler.cc — Phase 1: timer + SIGPROF + bump-buffer plumbing.
+ *
+ * Phase 1 scope:
+ *   - Allocate a single process-wide ring buffer (bump pointer, drop-on-full).
+ *   - Install an async-signal-safe SIGPROF handler that records a sample
+ *     header plus only the interrupted RIP. Frame-pointer walking is a
+ *     Phase 2 addition.
+ *   - Drive ITIMER_PROF at the requested rate. Portable between Linux and
+ *     macOS — both support setitimer(ITIMER_PROF).
```
Bike, line 10: `setitimer` is POSIX but has been marked obsolescent as of 2008 in favor of a newer timer API

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,487 @@
+/*
+ * sampling_profiler.cc — Phase 1: timer + SIGPROF + bump-buffer plumbing.
+ *
+ * Phase 1 scope:
+ *   - Allocate a single process-wide ring buffer (bump pointer, drop-on-full).
+ *   - Install an async-signal-safe SIGPROF handler that records a sample
+ *     header plus only the interrupted RIP. Frame-pointer walking is a
+ *     Phase 2 addition.
```
Bike, line 8: These "phases" are unclear. I guess they are phases of support being added? That's not really appropriate for comments.

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,487 @@
...
+std::atomic<uint64_t> g_samples_recorded{0};
+std::atomic<uint64_t> g_samples_dropped{0};
+
+struct sigaction      g_prev_sigaction;     // clasp's original SIGPROF handler
+bool                  g_prev_sigaction_saved = false;
+
+std::mutex            g_lifecycle_lock;     // serializes start/stop/save/reset
+
+// Read CLOCK_MONOTONIC in nanoseconds. clock_gettime is async-signal-safe
+// on Linux and macOS for CLOCK_MONOTONIC.
```
Bike, line 80: clock_gettime is specified to be async-signal-safe in POSIX, regardless of what clock you're using as far as I can see

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,487 @@
...
+static inline uint64_t now_ns_signal_safe() {
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  return (uint64_t)ts.tv_sec * 1000000000ull + (uint64_t)ts.tv_nsec;
+}
+
+// Read the interrupted instruction pointer out of the context structure.
+// x86_64 only for now — portable to arm64 when we need it.
+static inline uintptr_t ucontext_rip(void* ucptr) {
```
Bike, line 89: can we use libunwind or something instead of directly and nonportably accessing into an mcontext

include/clasp/core/sampling_profiler.h
```diff
@@ -0,0 +1,65 @@
...
+  uint64_t timestamp_ns;   // CLOCK_MONOTONIC at signal delivery
+  uint64_t vm_pc;          // bytecode VM's _pc at sample time, or 0
+  uint32_t thread_id;      // Linux tid / macOS port id (truncated)
+  uint32_t depth;          // number of trailing PCs (0 if walk failed)
+};
+
+// Start the profiler.
+//   rate_hz          : sampling rate in Hz (e.g. 97). Clamped to [1, 10000].
+//   max_depth        : per-sample stack-depth cap. Clamped to [1, 8192].
+//   buffer_bytes     : ring buffer size (0 = default 256 MiB).
```
Bike, line 49: we could just have the default be the actual number for 256 MiB rather than giving 0 a special meaning

src/gctools/snapshotSaveLoad.cc
```diff
@@ -570,6 +571,12 @@ void encodeEntryPoint(Fixup* fixup, uintptr_t* ptrptr, core::T_sp codebase, core
 #endif
   } else if (gc::IsA<core::BytecodeModule_sp>(codebase)) {
     encodeEntryPointInLibrary(fixup, ptrptr,"BytecodeModule");
+  } else if (codebase.nilp()) {
+    // No codebase object attached. Used by GFBytecodeSimpleFun where _Code is
```
Bike, line 575: we should use some object other than nil to indicate that we have a GFBytecodeSimpleFun. If there's no sensible code object it could just be an otherwise meaningless marker of some kind, as long as it specifically indicates GF bytecode

src/gctools/snapshotSaveLoad.cc
```diff
@@ -1069,11 +1078,20 @@ struct calculate_size_t {
       gctools::clasp_ptr_t client = HEADER_PTR_TO_GENERAL_PTR(header);
       size_t objectSize;
       if (header->_badge_stamp_wtag_mtag._value == DO_SHIFT_STAMP(gctools::STAMPWTAG_llvmo__ObjectFile_O)) {
+        llvmo::ObjectFile_O* code = (llvmo::ObjectFile_O*)client;
+        // Transient scaffolding (arena init's shared trampoline / stub
+        // template) must never appear in the snapshot; these ObjectFiles lack
+        // the standard __clasp_literals_<id> / function-vector-<id> symbols
+        // and would fail force_materialize at load time. Defense-in-depth —
+        // these are normally unreachable because they skip _AllObjectFiles
```
Bike, line 1086: there was a whole comment above about how they **don't** skip that registration

If this is referring the part below where they are removed from _AllObjectFiles before snapshot save, that should be made clearer

src/gctools/snapshotSaveLoad.cc
```diff
@@ -1945,6 +1968,35 @@ void snapshot_save(SaveLispAndDie& data) {
...
+          && gc::As_unsafe<llvmo::ObjectFile_sp>(car)->_TransientSkipSnapshot) {
+        ++dropped;
+      } else {
+        kept = core::Cons_O::create(car, kept);
+      }
+      cur = CONS_CDR(gc::As_unsafe<core::Cons_sp>(cur));
+    }
+    _lisp->_Roots._AllObjectFiles.store(kept, std::memory_order_relaxed);
+    if (dropped) {
+      fprintf(stderr, "[trampoline-arena] snapshot save dropped %zu transient ObjectFile(s) from _AllObjectFiles\n", dropped);
```
Bike, line 1994: unconditional debug output

src/llvmo/trampoline_arena.cc
```diff
@@ -0,0 +1,295 @@
...
+ * slots and a single memcpy is all that's required.
+ *
+ * For unwinding, each slot registers its own CIE+FDE with libgcc via
+ * __register_frame(slot + code_size). The trailing 4-byte zero terminator
+ * in the slot stops libgcc's walk after that one FDE.
+ *
+ * Memory: each page is mmap'd PROT_READ|PROT_WRITE|PROT_EXEC and never
+ * reprotected. Works on Linux (default policy allows RWX). On macOS arm64
+ * (W^X enforced) this needs MAP_JIT + per-thread pthread_jit_write_protect_np
+ * toggling, OR a dual-mapped scheme via shm_open.
```
Bike, line 23: If read-write shm_open on an object that's also executable is even possible, that's surely an OS security hole and not something we want to exploit.

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+
+enum class TrampolineKind { Bytecode, GF };
+
+namespace {
+
+static const char* kind_label(TrampolineKind k) {
+  return k == TrampolineKind::Bytecode ? "bytecode" : "GF";
+}
+
+static std::string build_trampoline_ir(uint64_t target_addr,
```
Bike, line 181: why do we build the IR as text rather than through IRBuilder

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  ir += "\n";
+  // Mirror the legacy trampoline IR's literals stub. clasp's
+  // ObjectFile-on-load machinery expects every JIT'd module to define it.
+  ir += "@__clasp_literals_trampoline_stub = internal local_unnamed_addr global [0 x ptr] zeroinitializer, align 8\n\n";
+  // No personality. The trampoline is a pure passthrough with no landing
+  // pads; the C++ unwinder only needs CFI to step through this frame, not
+  // a personality callback. Keeping personality off means LLVM emits a
+  // plain "zR" CIE whose only aug-data byte is the FDE encoding, with no
+  // pcrel-encoded personality pointer — so the CIE bytes are fully
+  // position-independent and can be memcpy'd verbatim into every arena
```
Bike, line 213: I don't actually know what all this CIE and FDE stuff means, but it again sounds like we're relying on LLVM to generate position-independent DWARF based on this superficially unrelated question of whether there's a personality.

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  const char* closure_param = gf ? "%0" : "%1";
+  const char* nargs_param   = gf ? "%1" : "%2";
+  const char* args_param    = gf ? "%2" : "%3";
+
+  std::string ir;
+  ir.reserve(2048);
+  ir += targets;
+  ir += "\n";
+  // Mirror the legacy trampoline IR's literals stub. clasp's
+  // ObjectFile-on-load machinery expects every JIT'd module to define it.
```
Bike, line 206: This is true but it might be better to alter or eliminate that machinery for trampolines that do not need it.

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  // plain "zR" CIE whose only aug-data byte is the FDE encoding, with no
+  // pcrel-encoded personality pointer — so the CIE bytes are fully
+  // position-independent and can be memcpy'd verbatim into every arena
+  // slot.
+  //
+  // Direct inttoptr-in-call: forces LLVM to materialize the address as a
+  // 64-bit immediate (movabs on x86_64; MOVZ+MOVK*3 on arm64). The bytes
+  // generated are byte-identical for every compile (the immediate is the
+  // same value across all slots), so memcpy alone is enough — no per-slot
+  // patching, no RIP-relative offsets to fix up.
```
Bike, line 220: trying to force LLVM to generate something pretty particular by doing hacky instructions. If LLVM actually documents that it generates code that doesn't need linking if you do enough pointer arithmetic, we can point at that documentation in the comment. Otherwise this is working by coincidence.

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  //
+  // uwtable: emit unwind tables so Lisp Unwind exceptions can propagate.
+  // frame-pointer=all: keep the frame pointer chain intact for backtrace.
+  ir += "attributes #0 = { uwtable \"frame-pointer\"=\"all\" }\n";
+  return ir;
+}
+
+// Walk an x86_64-ELF .eh_frame section and extract the single CIE and the
+// single FDE as verbatim byte blobs. Relies on the trampoline IR having
+// exactly one uwtable function (the template), with the _end marker
```
Bike, line 258: is this the same _end marker that the above comment says doesn't exist

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+// attributed as no-uwtable — so LLVM emits one CIE + one FDE. Any other
+// count (zero, or two+ of either) is a sign that something upstream
+// changed and silent copy-of-wrong-bytes is about to happen: we fail loud.
```
Bike, line 261: "a sign that something upstream changed". This is an admission that this code relies on undocumented behavior of libraries

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+//
+// Structural bytes (length, CIE pointer, aug data, PC range, CFI opcodes)
+// need no relocations, so what's in the unlinked ObjectFile is already
+// correct. The only field that requires a relocation is the FDE's PC
+// begin; its pre-relocation value is typically zero, and the caller
+// patches it for the slot layout.
```
Bike, line 267: so we ARE patching the bytes llvm gives us, or what?

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  uint8_t aug_len = cie[pos++];
+  if (aug_len != 1) {
+    fprintf(stderr,
+            "[trampoline-arena] CIE aug_length = %u (expected 1 for \"zR\")\n",
+            aug_len);
+    return false;
+  }
+  if (pos >= n) return false;
+  uint8_t fde_enc = cie[pos];
+  // DW_EH_PE_pcrel = 0x10, DW_EH_PE_sdata4 = 0x0B. Combined = 0x1B.
```
Bike, line 392: so it looks like what we're doing is having LLVM generate the DWARF, but then failing if it generates anything other than a specific sequence of bytes?

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  constexpr uint8_t kExpectedFdeEncoding = 0x1B;
+  if (fde_enc != kExpectedFdeEncoding) {
+    fprintf(stderr,
+            "[trampoline-arena] CIE FDE encoding = 0x%02x (expected 0x%02x = "
+            "pcrel|sdata4). LLVM may have switched .eh_frame defaults — the "
```
Bike, line 397: another reference to libraries (LLVM) changing the undocumented behavior we're relying on

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,866 @@
...
+  // LSDA) and the FDE PC encoding being pcrel|sdata4. If either drifts
+  // out from under us, the byte-identical memcpy silently corrupts unwind
+  // info. Fail at install time instead.
+  if (!validate_trampoline_cie(cie_bytes.data(), cie_bytes.size())) {
+    state.store(2, std::memory_order_release);
+    return false;
+  }
+
+  // Patch the FDE's PC begin field for the slot layout
+  // [code | CIE | FDE | terminator]. The CIE uses DW_EH_PE_pcrel|sdata4
```
Bike, line 674: yep, we're relying on specific bytes again

Bike (commented):
still has plenty of hardcoding, just more subtly

PR #1769, opened 2026-05-24
Add core:build-mode to report bytecode vs native build at runtime

dg1sbg:
## Summary

Adds `core:build-mode`, a runtime introspection function returning `:bytecode` or `:native` — the koga `build-mode` the running image was compiled with.

Clasp already records the GC, debug/release, threads, float configs, etc. in `*features*`, but the bytecode-vs-native build mode was only known at build time (it sets `STARTUP_IMAGE_EXTENSION`); there was no way to query it at runtime. This adds that.

## Changes

- **`src/koga/config-header.lisp`**: emit a `CLASP_BUILD_MODE_BYTECODE` define in `config.h` for bytecode builds. Mirrors the existing `USE_BOEHM`/`USE_MMTK` idiom — `write-defines` omits it entirely for native builds, so `#ifdef` is the natural test.
- **`src/core/primitives.cc`**: add `core:build-mode` (`CL_DEFUN`) returning the interned keyword via `#ifdef`, next to the other `lisp-implementation-*` introspection functions. No symbol-table/export edits needed; the scraper handles interning/exposure like its siblings.

The docstring is explicit that this reports the **build-time configuration**, not what `compile` currently emits — a `:native` build still carries a bytecode compiler, and a `:bytecode` build can JIT to native via auto-compilation.

## Verification

Built and ran the `boehmprecise` variant in both modes (macOS/Apple Silicon, LLVM 22):

- native build → `(core:build-mode)` returns `:NATIVE`
- bytecode-configured build (`config.h` defines `CLASP_BUILD_MODE_BYTECODE`) → `(core:build-mode)` returns `:BYTECODE`

`(fboundp 'core:build-mode)` returns `T`, and the symbol reads as `core:build-mode` (exported from `CORE`).

### DIFF
--- a/src/core/primitives.cc
+++ b/src/core/primitives.cc
@@ -269,6 +269,18 @@ CL_DEFUN T_sp core__clasp_git_full_commit() {
 #endif
 };
 
+CL_LAMBDA();
+CL_DECLARE();
+CL_DOCSTRING(R"dx(build-mode - :bytecode or :native, the mode clasp's image was built with. Note: this reports the build-time configuration, not what COMPILE currently emits - a :native build still carries a bytecode compiler, and a :bytecode build can JIT to native via auto-compilation.)dx");
+DOCGROUP(clasp);
+CL_DEFUN T_sp core__build_mode() {
+#ifdef CLASP_BUILD_MODE_BYTECODE
+  return _lisp->internKeyword("BYTECODE");
+#else
+  return _lisp->internKeyword("NATIVE");
+#endif
+};
+
 CL_LAMBDA(obj);
 CL_DECLARE();
 CL_DOCSTRING(
--- a/src/koga/config-header.lisp
+++ b/src/koga/config-header.lisp
@@ -25,6 +25,7 @@
                  (if (eq (build-mode configuration) :bytecode)
                      "fasl"
                      "nfasl")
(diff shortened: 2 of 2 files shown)


Bike:
Why do you want this? And please rebase against main, tests are failing because of some problems that have since been fixed.

dg1sbg:
> Why do you want this? And please rebase against main, tests are failing because of some problems that have since been fixed.

I need to easily and reliably determine at runtime which mode clasp was built in. This is for using clasp on an embedded platform (Analog Devices ADSP-SC598, ARM-based) and making sure that we are able to compile stuff on that embedded platform (even using #'compile while clasp is executed on that DSP processor). The app will an audio processor. 

This was under the assumption that a bytecode-built clasp will also only be compiling to bytecode when using #'compile.

Or is there another way I just am not aware of?

And yes, I will rebase, of course. Thanks for the feedback.

Bike:
Right, I thought that might be it, which is why I asked. I've been trying to disentangle the build mode from whether the built image compiles natively or not. Whether `compile-file` does native compilation is controlled by the `:native` argument which defaults to `cmp:*compile-file-native*`, the initial value of which is given by the `:default-native` option to koga. `compile` currently always native compiles regardless of build mode. I'll put in a control for that, also defaulting to `:default-native`: #1789.

dg1sbg:
Thanks - so I close this request.

PR #1770, opened 2026-05-25
Harden the bytecode FASL linker against transient/failed syscalls

dg1sbg:
## Summary

`core__link_fasl_files` / `link_fasl_files` (the bytecode image linker in `src/core/loadltv.cc`) turned any failed syscall into a fatal `SIMPLE_ERROR`, and never checked `open()`. Under a heavily parallel build, a transient syscall failure could abort the whole build at the `Linking … base.fasl` step.

The unchecked `open()` was especially confusing: on failure it returns −1, after which `lseek`/`mmap` operate on a bad descriptor, so the error surfaced as a misleading `Could not mmap … Bad file descriptor` instead of the real `open()` error.

## Changes (`src/core/loadltv.cc`)

- **Check `open()`** and report the real `errno`, instead of a bogus downstream `mmap` error.
- **Retry transient syscall failures** — `open`/`mkstemp`/`munmap`/`rename`/`close` on `EINTR` (clasp's GC and thread-coordination signals can interrupt syscalls), with bounded backoff on `EMFILE`/`ENFILE`/`EAGAIN`. `mkstemp` restores its `XXXXXX` template before each retry.
- **`write_all()`** — complete the write across partial writes and `EINTR`; a bare `write()` could silently truncate the linked FASL.

The success path is unchanged; this only makes transient failures recoverable and diagnostics accurate, and fixes two latent bugs (unchecked `open()`, and silent truncation from an unchecked `write()`).

## Verification

Found via a `--build-mode=bytecode` build on macOS / Apple Silicon (LLVM 22) that intermittently aborted at the FASL-link step. With this change, a clean from-scratch bytecode build completes end-to-end — links `base.fasl`, boots `iclasp` from it, and builds the modules — and a 48-way parallel link stress passes 0/48.

Note: the original abort was a rare transient (not deterministically reproducible in isolation), so this is defensive hardening of the syscalls the crash backtrace implicated, plus genuine fixes for the unchecked `open()`/`write()`.

### DIFF
--- a/src/core/loadltv.cc
+++ b/src/core/loadltv.cc
@@ -1315,6 +1315,71 @@ struct ltv_MmapInfo {
   ltv_MmapInfo(uint8_t* mem, size_t len) : _Memory(mem), _Len(len) {};
 };
 
+// Apple Silicon / parallel-build hardening for the FASL linker.
+//
+// clasp's GC and thread-coordination signals can interrupt slow syscalls
+// (EINTR), and a heavily parallel build can momentarily run into file-descriptor
+// pressure (EMFILE/ENFILE). The original code retried none of these and did not
+// check open(), so a transiently-failed open() returned -1 and only surfaced
+// later as a fatal and misleading "Could not mmap ... Bad file descriptor"
+// abort, breaking the whole build. These helpers retry transient failures so the
+// linker is robust under load.
+static bool ltv_transient_errno() {
+  return errno == EINTR || errno == EMFILE || errno == ENFILE || errno == EAGAIN;
+}
+
+static int ltv_retry_open(const char* path, int flags) {
+  for (int attempt = 0;; ++attempt) {
+    int fd = ::open(path, flags);
+    if (fd >= 0 || !ltv_transient_errno() || attempt >= 1000)
+      return fd;
+    if (errno != EINTR)
+      usleep(1000); // brief backoff for fd-pressure (EMFILE/ENFILE/EAGAIN)
+  }
+}
+
+static int ltv_retry_close(int fd) {
+  int r;
+  do {
+    r = ::close(fd);
+  } while (r < 0 && errno == EINTR);
+  return r;
+}
+
+static int ltv_retry_munmap(void* addr, size_t len) {
+  int r;
+  do {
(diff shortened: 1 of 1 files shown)


src/core/loadltv.cc
```diff
@@ -1315,14 +1315,89 @@ struct ltv_MmapInfo {
...
+  do {
+    r = ::close(fd);
+  } while (r < 0 && errno == EINTR);
+  return r;
+}
+
+static int ltv_retry_munmap(void* addr, size_t len) {
+  int r;
+  do {
+    r = ::munmap(addr, len);
```
Bike, line 1352: Linux and Mac man pages do not list `munmap` as ever returning EINTR

src/core/loadltv.cc
```diff
@@ -1315,14 +1315,89 @@ struct ltv_MmapInfo {
   ltv_MmapInfo(uint8_t* mem, size_t len) : _Memory(mem), _Len(len) {};
 };
 
+// Apple Silicon / parallel-build hardening for the FASL linker.
+//
+// clasp's GC and thread-coordination signals can interrupt slow syscalls
+// (EINTR), and a heavily parallel build can momentarily run into file-descriptor
+// pressure (EMFILE/ENFILE). The original code retried none of these and did not
```
Bike, line 1322: no need for comments to describe the original code - leave that for commit messages

src/core/loadltv.cc
dg1sbg: Source code from Linux kernel (several levels down in the code):

```
int __vm_munmap(unsigned long start, size_t len, bool unlock)
{
	int ret;
	struct mm_struct *mm = current->mm;
	LIST_HEAD(uf);
	VMA_ITERATOR(vmi, mm, start);

	if (mmap_write_lock_killable(mm))
		return -EINTR;  // *** HERE ***

	ret = do_vmi_munmap(&vmi, mm, start, len, &uf, unlock);
	if (ret || !unlock)
		mmap_write_unlock(mm);

	userfaultfd_unmap_complete(mm, &uf);
	return ret;
}
```
The man pages don't say it, but it's there and I do get EINTR return values.

PR #1771, opened 2026-05-25
Runtime performance optimizations + Apple Silicon FFI-callback W^X fix

dg1sbg:
## Summary

A set of independent runtime performance improvements (each its own commit), plus one Apple-Silicon correctness fix. All changes are portable — the platform-specific bits compile to no-ops off Apple Silicon.

| commit | change | measured effect |
|---|---|---|
| Default `*print-pretty*` to NIL | clasp's pretty printer is a CLOS Gray stream (C++↔Lisp dispatch per character); defaulting it on made every `princ`/`prin1`/`format ~A`/`print` pay that cost | **~370×** faster non-pretty printing (a 50-elt list: ~2.4 ms → ~6 µs). REPL stays pretty (`tpl-print` rebinds it); `pprint`/`~<~:>`/explicit bindings unaffected |
| Non-atomic per-thread alloc counters | the counters live in the `THREAD_LOCAL` `ThreadLocalStateLowLevel` and are never accessed cross-thread, so the `std::atomic`s were pure overhead on every allocation (also fixes 3 uninitialized fields) | 2–6% on allocation |
| Bulk `StringOutputStream_O::write_string` | replaces the per-character boxing + virtual `vectorPushExtend` with a single grow + typed bulk copy + scan-once cursor; safe fallback for narrowing | **1.8×–62×** (scales with length), output byte-identical |
| `countObjectFileNames` O(N²)→O(1) | it rescanned `_AllObjectFiles` per JIT-module registration; replaced with a mutex-guarded `name→count` index synced at the single add point + the bulk-clear sites | removed the #1 self-time function (~15%) in JIT-heavy workloads |
| Enable Boehm `PARALLEL_MARK` | it was off (config-header template default) although bdwgc defaults it on and `GC_THREADS`/`THREAD_LOCAL_ALLOC` are enabled; clasp marks conservatively with no custom mark procs, so parallel markers are safe | **~3×** faster GC marking on multicore |
| Cache the TLS pointer in the bytecode VM | on Darwin each `thread_local` access is a `_tlv_get_addr` thunk; the interpreter hit `my_thread` per call | removes the per-call thunk (no-op load off Darwin) |
| **Fix Apple-Silicon W^X SIGBUS in FFI callbacks** | `setf_jit_lookup_t` stored a function pointer into JIT'd (MAP_JIT) callback memory without enabling writes — `%defcallback` crashed at compile time. Wrapped with the existing `JITDataReadWriteMaybeExecute()`/`JITDataReadExecute()` helpers | `defcallback` now works (was a hard SIGBUS) |

## Testing

Built and tested on macOS Apple Silicon (LLVM 22), `boehmprecise` variant:

- **Full regression suite: 1877 pass, 0 bus errors / segfaults**; the only remaining failures are the pre-existing known ones in `set-unexpected-failures.lisp` (sbcl-cross-compile, include-level, types-classes).
- The W^X fix turns the previously-crashing `defcallback-native` test into a pass (`CFFI-DEFCALLBACK` — the Lisp callback drives C `qsort` correctly).
- Output correctness verified byte-identical against a base/extended-string, tab/newline, fill-pointer-growth, and narrowing golden test.
- Parallel marking verified stable: full regression byte-identical across 3 runs (no non-determinism).

Commits are independent and can be reviewed/cherry-picked separately. The only user-visible behavior change is the `*print-pretty*` default (deliberate, matching SBCL/CCL/CLISP).

### DIFF
--- a/include/clasp/core/lispStream.h
+++ b/include/clasp/core/lispStream.h
@@ -685,6 +685,9 @@ public:
   String_sp get_string();
 
   claspCharacter write_char(claspCharacter c) override;
+  // Bulk override: avoids the per-character boxing + virtual vectorPushExtend
+  // of the default AnsiStream_O::write_string char-by-char loop.
+  void write_string(String_sp data, cl_index start, cl_index end) override;
   void clear_output() override;
   void finish_output() override;
   void force_output() override;
--- a/include/clasp/gctools/boehm_config.h
+++ b/include/clasp/gctools/boehm_config.h
@@ -205,7 +205,14 @@
 #define PACKAGE_VERSION "8.2.8"
 
 /* Define to enable parallel marking. */
-/* #undef PARALLEL_MARK */
+/* clasp: ENABLED. Heap marking is Boehm-standard conservative marking
+   (ALL_INTERIOR_POINTERS=1 + GC_register_displacement for tagged pointers; clasp
+   registers NO custom mark procedures / typed descriptors), so parallel markers
+   parallelize Boehm's own, parallel-safe marking loop -- clasp injects no
+   per-thread marking logic. GC_THREADS and THREAD_LOCAL_ALLOC are already on, and
+   bdwgc auto-starts (#CPUs-1) marker threads in GC_thr_init. This parallelizes
+   GC_mark_from, the dominant runtime cost (~25-30% in profiles) on multicore. */
+#define PARALLEL_MARK 1
 
 /* If defined, redirect free to this function. */
 /* #undef REDIRECT_FREE */
--- a/include/clasp/gctools/threadlocal.fwd.h
+++ b/include/clasp/gctools/threadlocal.fwd.h
@@ -25,13 +25,21 @@ struct MonitorAllocations {
 #endif
 
 struct GlobalAllocationProfiler {
-  std::atomic<int64_t> _BytesAllocated;
+  // These counters live in the THREAD_LOCAL ThreadLocalStateLowLevel (as the
+  // member _Allocations) and are only ever accessed via
+  // my_thread_low_level->_Allocations, i.e. by the owning thread alone (see
+  // gcalloc_boehm.h, gcFunctions.cc, startRunStop.cc, memoryManagement.cc).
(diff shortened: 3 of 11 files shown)


Bike:
Moved most of the optimization into #1791 as suggested.

The `*print-pretty*` thing I'm not doing since you can always configure it yourself in .clasprc or what have you, and making things look nice by default is good. Also, SBCL actually defaults to `*print-pretty*` being true, so I don't appreciate that little bit of misinformation. Presumably Claude sees the defvar in the SBCL source and not the way it's set in cold init.

Boehm parallel mark is probably causing that CI failure above (in the Ubuntu build, Mac has other problems), so that needs fixing.

The Mac fixes I'm going to put together into their own PR.

Bike:
The JIT W^X part appears to have been entirely duplicated by #1786, and I think it makes more sense to separate it out from the rest of this. So the only remaining part of this PR is the parallel marking.

dg1sbg:
Closing: the content of this PR has already landed on `main` under other commits, so there is nothing left to merge.

Checked while building an integration branch of my open PRs. This branch's merge-base with `main` is `f42255947` (27 Apr) and `main` is now **138 commits ahead**, producing 10 conflict hunks across 5 files. Every one of those conflicts turned out to be the *same* optimization already present on `main`, differing only in comment wording:

| optimization in this PR | already on `main`? |
|---|---|
| `bytecode.cc` — cache the TLS pointer once per VM frame | yes |
| `code.cc` — O(1) `objectFileNameCounts` index replacing the O(N²) rescan | yes |
| `lispStream.cc` — single geometric growth for bulk string output | yes |
| `llvmoPackage.cc` — W^X write-enable around the JIT'd global store | yes |

The only content not upstream is the `AllocationProfiler` → `GlobalAllocationProfiler` rename with atomics replaced by plain `int64_t`. That is small enough to raise separately if it is still wanted, and it would need writing against current `main` rather than rebasing 138 commits of drift.

Rebasing this branch would produce an essentially empty diff plus that rename, so closing is cleaner than carrying it. Reopen or ping me if you would rather I rebase it anyway.

PR #1775, opened 2026-05-26
Boot speedup: buffer FASL reads + cache file-scope (~2.6x faster boot)

dg1sbg:
## Summary

Two small, independent loader changes that more than halve clasp's boot time. Both touch only `src/core/loadltv.cc` and are portable C++ (no platform specifics).

### 1. Buffer FASL reads in `loadltv`
`loadltv` was reading the FASL one byte / one small field at a time straight from the stream — `read_u8()` → `stream_read_byte()`, and the multi-byte/blob reads → `stream_read_byte8()`. Every such call is a virtual stream dispatch plus an interrupt-park (`BEGIN_PARK`/`END_PARK`). The base image is loaded entirely through this path, so boot was dominated by per-byte stream overhead.

Add a 64 KB read-ahead buffer and route **every** read through it: `read_u8`, `read_u16/32/64/80/128`, the base-string / utf8-string / machine-code blob reads, the header, and the bytecode-module read that was using `cl__read_sequence`. The loader owns the stream for the whole load (`load_bytecode_stream` makes a fresh loader and the caller closes the stream afterwards), so reading ahead is safe; it also now errors cleanly on unexpected EOF instead of reading uninitialized data (which closes a latent issue).

### 2. Cache the last (path → FileScope) in `di_op_location`
After buffering, the next-biggest chunk of boot was constructing debug-info locations. `di_op_location` calls `core__file_scope(path)` per location — that copies the string and does a locked hash intern. The base image has many locations per source file that share the same path literal, so this was redundant.

Cache the last (path-literal → FileScope) in the loader and reuse it on an `eq` match, falling back to `core__file_scope` on a miss — so correctness holds even without literal coalescing.

## Measured (on Apple Silicon, macOS, LLVM 22, boehmprecise variant)

`time clasp --noinform --non-interactive -e '(core:exit)'`, best of 8 runs:

| | boot wall | `read()` sys time |
|---|---|---|
| baseline | 5.14 s | 2.02 s |
| + buffered reads | ~2.20 s | 0.12 s |
| + file-scope cache | **~1.95 s** (min 1.94 s) | — |

**~2.6× faster boot.** The 16× drop in `read()` sys time is the direct, machine-independent proof that the per-byte I/O overhead is gone.

## Testing

Full clasp regression suite **byte-identical** to the pre-change baseline after each commit: **1877 pass, 0 load errors / bus errors / segfaults**. (The 4 remaining "failures" are the pre-existing known issues listed in `set-unexpected-failures.lisp` — unrelated.) Because the suite compiles and loads thousands of FASLs through the changed path, the buffered parser and the file-scope cache are well-exercised.

### DIFF
--- a/src/core/loadltv.cc
+++ b/src/core/loadltv.cc
@@ -82,9 +82,51 @@ struct loadltv {
   // native modules
   T_sp _JITDylib = nil<T_O>();
 
-  loadltv(Stream_sp stream) : _stream(stream) {}
+  // Read buffer. loadltv otherwise pulls the FASL one byte / one small field at
+  // a time straight from the stream, and every such call is a virtual stream
+  // dispatch plus an interrupt-park (BEGIN_PARK/END_PARK). Reading the stream in
+  // large chunks and serving bytes from this buffer removes that per-byte
+  // overhead, which dominates boot (the base image is loaded through this path).
+  // The loader owns the stream for the whole load (load_bytecode_stream makes a
+  // fresh loader and the caller closes the stream afterwards), so reading ahead
+  // is safe.
+  static constexpr size_t BUFSIZE = 65536;
+  std::vector<unsigned char> _buf;
+  size_t _bufpos = 0;
+  size_t _buflen = 0;
 
-  inline uint8_t read_u8() { return stream_read_byte(_stream).unsafe_fixnum(); }
+  loadltv(Stream_sp stream) : _stream(stream), _buf(BUFSIZE) {}
+
+  // Refill the buffer from the stream; returns the number of bytes now available.
+  size_t refill() {
+    _buflen = (size_t)stream_read_byte8(_stream, _buf.data(), BUFSIZE);
+    _bufpos = 0;
+    return _buflen;
+  }
+
+  // Copy n bytes of the (buffered) stream into dst.
+  inline void read_bytes(unsigned char* dst, size_t n) {
+    while (n > 0) {
+      if (_bufpos >= _buflen) {
+        if (refill() == 0)
+          SIMPLE_ERROR("Invalid FASL: unexpected end of stream");
+      }
+      size_t avail = _buflen - _bufpos;
+      size_t take = (n < avail) ? n : avail;
+      memcpy(dst, _buf.data() + _bufpos, take);
+      _bufpos += take;
+      dst += take;
+      n -= take;
+    }
+  }
+
+  inline uint8_t read_u8() {
+    if (_bufpos >= _buflen) {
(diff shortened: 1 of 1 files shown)


src/core/loadltv.cc
```diff
@@ -81,10 +81,58 @@ struct loadltv {
...
   // native modules
   T_sp _JITDylib = nil<T_O>();
+  // Cache the last (debug-info path literal -> FileScope). Consecutive debug
+  // locations from the same source file share the same path literal, so this
+  // avoids a redundant string copy + locked hash intern (core__file_scope) per
+  // location during boot. eq-compared; falls back to the real call on a miss.
+  T_sp _lastDebugPath = nil<T_O>();
+  T_sp _lastDebugScope = nil<T_O>();
+
+  // Read buffer. loadltv otherwise pulls the FASL one byte / one small field at
```
Bike, line 91: again, describing what used to happen but no longer happens is good for a commit message, not for comments.

src/core/loadltv.cc
```diff
@@ -734,7 +782,10 @@ struct loadltv {
     size_t index = next_index();
     uint32_t len = read_u32();
     SimpleVector_byte8_t_sp bytes = SimpleVector_byte8_t_O::make(len);
-    cl__read_sequence(bytes, _stream, clasp_make_fixnum(0), nil<T_O>());
+    // Read the module bytecode straight into the vector through our buffer
+    // (equivalent to the previous cl__read_sequence, but without per-call stream
```
Bike, line 786: please fix this comment as well

src/core/loadltv.cc
```diff
@@ -964,8 +1015,14 @@ struct loadltv {
     Integer_sp end = Integer_O::create(read_u32());
     T_sp path = get_ltv(read_index());
     uint64_t line = read_u64(), column = read_u64(), filepos = read_u64();
-    T_mv sfi_mv = core__file_scope(path);
-    FileScope_sp sfi = gc::As<FileScope_sp>(sfi_mv);
+    FileScope_sp sfi;
+    if (path == _lastDebugPath && _lastDebugScope.notnilp()) {
+      sfi = gc::As_unsafe<FileScope_sp>(_lastDebugScope);
```
Bike, line 1020: We're moving to use T_sp functions instead of `gc::As` - this should be `_lastDebugScope.as_unsafe<FileScope_O>()`. and probably `as_assert` rather than `as_unsafe` but that's no big deal.

Bike (commented):
Looks good other than the small stuff!

PR #1776, opened 2026-05-26
Security hardening: bound FASL sizes, refuse unsafe socket buffers, shell-quote save-lisp paths (S2–S10 except S1/S6)

dg1sbg:
## Summary

Six small, file-local security fixes addressing pre-existing vulnerabilities at the FASL loader, sockets FFI, save-lisp-and-die build path, exec-environment construction, and the LLVM symbol-lookup callback. Each commit is independent and reviewable on its own.

### Threat-model framing

clasp has no internal security boundary by design — loading a FASL is arbitrary code execution, and any Lisp code already has full `ext:system` / FFI. These fixes therefore matter at real trust boundaries: loading FASLs from a less-trusted source, feeding network bytes to the socket primitives, or passing externally-influenced paths to `save-lisp-and-die`.

### The fixes

| commit | issue | fix |
|---|---|---|
| `loadltv: validate FASL-supplied bignum and array sizes` | (S2) `op_bignum` did `mp_limb_t limbs[std::abs(ssize)]` — stack VLA sized from a file byte, with `std::abs(INT64_MIN)` UB and no upper bound. (S7) `op_array`'s `total *= dim` could overflow `size_t` then be passed to `make_vector`/`make_mdarray` while the per-element fill walked the full unwrapped count → read/write past the buffer | reject `INT64_MIN`, cap limbs at `1<<20`, heap-allocate via `std::vector`; replace `total *= dim` with `__builtin_mul_overflow` and cap at `1<<30` elements |
| `sockets: refuse unsafe buffer types + range-check fd_set fds` | (S3/S4) `safe_buffer_pointer`'s `(vector t)` branch let `recvfrom`/`send` scribble raw network bytes over a vector of *boxed Lisp pointers* (GC heap corruption); its `(size+divisor-1)/divisor` capacity check wrapped on a negative `length` cast to `uint`. (S8) `FD_SET`/`FD_CLR`/`FD_ISSET` from Lisp had no range check, writing/reading out of bounds for `fd >= FD_SETSIZE` | refuse anything other than strings or `(simple-array (unsigned-byte 8))`, reject negative size, change the param type to `int`; range-check fd in all three operators |
| `serve-event: range-check fd_set fds` | (S8) same `FD_SET`/`FD_ISSET` issue in serve-event's variants | range-check fd |
| `save-lisp-and-die: shell-quote interpolated paths` | (S5) `system(...)` and `popen(...)` built command lines by string concatenation with `filename`/`_FileName`/`_LibDir`. `(save-lisp-and-die :executable t :file "foo; rm -rf ~/work; #")` would run the `rm` at link time | small `shell_quote` helper (single-quote + escape internal `'`), wrap every user-controlled value before interpolation; build-time-constant `BUILD_LINKFLAGS`/`BUILD_LIB` stay unquoted, and `mangled_name` is already restricted to `[A-Za-z0-9_]` |
| `unixsys: terminate env strings before advancing the index` | (S9) `environ[j]=malloc(...); memcpy(...); j++; environ[j][l]=0;` — the NUL-terminate fired *after* `j++`, so it wrote into the next, still-uninitialized pointer slot at offset `l` (wild write) and the actual string was never terminated | move the terminate before `j++` |
| `Bound the LLVM symbol-lookup callback buffer copy` | (S10) `strcpy(global_LLVMSymbolLookupCallbackBuffer /* [1024] */, stringstream.str().c_str())` — buffer overflow when `dbg_safe_repr` emits a long object printout (JIT debug path) | bounded `memcpy` that truncates to `CALLBACK_BUFFER_SIZE-1` and NUL-terminates |

## What this PR does *not* do

- **(S1) No load-time bytecode verifier.** With `DEBUG_VIRTUAL_MACHINE` off (release), VM bounds checks are compiled out, so a malformed FASL can do anything the VM can do. Adding a verifier is a much larger, separate design project — out of scope here.
- **(S6) FASL loader EOF check** — already covered by clasp-developers/clasp#1775 (the buffered loader errors cleanly on unexpected EOF).

## Testing

Full clasp regression suite **byte-identical** to the pre-change baseline after the combined set: **1877 pass, 0 bus errors / segfaults / memory corruption**. (Same 4 pre-existing/expected failures listed in `set-unexpected-failures.lisp`, unrelated.) The suite exercises FASL loading, the sockets FFI, and `save-lisp-and-die` paths.

### DIFF
--- a/src/core/backtrace.cc
+++ b/src/core/backtrace.cc
@@ -116,12 +116,16 @@ static SimpleBaseString_sp lu_procname(unw_cursor_t* cursorp) {
       return SimpleBaseString_O::make("<unknown libunwind error>");
     }
     if (nbytes >= 4096) {
-      // Too long. Give up, putting an ellipsis on the end
-      char fnamedot[nbytes + 3];
-      fnamedot[0] = '\0';
-      strcat(fnamedot, fname);
-      strcat(fnamedot, "...");
-      return SimpleBaseString_O::make(fnamedot);
+      // Too long. libunwind returned UNW_ENOMEM, which means it filled (up to)
+      // nbytes bytes and may NOT have written a terminating NUL. The previous
+      // code did strcat(fnamedot, fname), which assumes fname is NUL-terminated
+      // -- without that guarantee it reads off the end of the buffer into the
+      // stack until it finds a NUL and writes that past fnamedot. Bound the
+      // read with strnlen and build the result without strcat.
+      size_t len = strnlen(fname, nbytes);
+      std::string s(fname, len);
+      s.append("...");
+      return SimpleBaseString_O::make(s);
     } else
       nbytes <<= 1;
   } while (true);
--- a/src/core/backtrace.cc
+++ b/src/core/backtrace.cc
@@ -117,12 +117,16 @@ static SimpleBaseString_sp lu_procname(unw_cursor_t* cursorp) {
       return SimpleBaseString_O::make("<unknown libunwind error>");
     }
     if (nbytes >= 4096) {
-      // Too long. Give up, putting an ellipsis on the end
-      char fnamedot[nbytes + 3];
-      fnamedot[0] = '\0';
-      strcat(fnamedot, fname);
-      strcat(fnamedot, "...");
-      return SimpleBaseString_O::make(fnamedot);
+      // Too long. libunwind returned UNW_ENOMEM, which means it filled (up to)
+      // nbytes bytes and may NOT have written a terminating NUL. The previous
+      // code did strcat(fnamedot, fname), which assumes fname is NUL-terminated
+      // -- without that guarantee it reads off the end of the buffer into the
+      // stack until it finds a NUL and writes that past fnamedot. Bound the
+      // read with strnlen and build the result without strcat.
+      size_t len = strnlen(fname, nbytes);
+      std::string s(fname, len);
+      s.append("...");
+      return SimpleBaseString_O::make(s);
     } else
       nbytes <<= 1;
   } while (true);
--- a/src/core/loadltv.cc
+++ b/src/core/loadltv.cc
@@ -571,8 +571,17 @@ struct loadltv {
     for (size_t i = 0; i < rank; ++i) {
       uint16_t dim = read_u16();
       dims << clasp_make_fixnum(dim);
-      total *= dim;
+      // Overflow-checked: with rank up to 255 and dim up to 0xFFFF, total can
+      // overflow size_t on a malformed FASL, then be passed to make_vector /
+      // make_mdarray as a wrapped-around small value -- mismatch with the per-
+      // element fill loop would read/write past the allocated buffer.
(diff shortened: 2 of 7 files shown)


dg1sbg:
Heads-up on a merge hazard in `src/gctools/snapshotSaveLoad.cc`, found while building an integration branch of the open PRs.

This PR's `shell_quote` changes collide with the aarch64/Darwin snapshot work that is also in flight, and **taking this PR's version of the two command constructions on its own is a regression** — not a textual conflict you can resolve by picking a side. Both halves are needed.

**1. objcopy invocation.** This PR hardcodes the x86-64 target:

```lisp
OBJCOPY_BINARY " --input-target binary --output-target elf64-x86-64"
               " --binary-architecture i386 "
```

On aarch64 those values produce `architecture i386 unknown` and an empty/rejected binary-wrapped snapshot object. The other side makes the target arch-conditional. The merged form keeps the arch conditional *and* this PR's `shell_quote`:

```lisp
#if defined(__x86_64__)
    const char* snapshot_objcopy_target = "elf64-x86-64";
    const char* snapshot_objcopy_arch = "i386";
#elif defined(__aarch64__)
    const char* snapshot_objcopy_target = "elf64-littleaarch64";
    const char* snapshot_objcopy_arch = "aarch64";
#endif
    cmd = OBJCOPY_BINARY " --input-target binary --output-target " +
          std::string(snapshot_objcopy_target) + " --binary-architecture " +
          snapshot_objcopy_arch + " " +
          shell_quote(filename) + " " + shell_quote(obj_filename) + ...
```

**2. Darwin link command.** This PR's version drops the absolute `-L` for the lib directory:

```lisp
cmd = CXX_BINARY " " BUILD_LINKFLAGS " -o" + shell_quote(snapshot_data->_FileName) + ...
```

`BUILD_LINKFLAGS` carries only a *relative* `-Lboehmprecise/lib`, so without an absolute `-L` the `-lclasp` in `save-lisp-and-die :executable` resolves only when run with `CWD=build/`. The Linux branch a few lines above already passes `-L" + shell_quote(snapshot_data->_LibDir)`; Darwin needs the same:

```lisp
cmd = CXX_BINARY " " BUILD_LINKFLAGS " -L" + shell_quote(snapshot_data->_LibDir) +
      " -o" + shell_quote(snapshot_data->_FileName) + ...
```

Note `shell_quote` **supersedes** the manual `\"` quoting the other side used — it single-quote-wraps and escapes embedded quotes, so it covers build paths containing spaces as well as the injection case this PR is about. No need to keep both quoting styles.

Verified with both halves combined: clean from-scratch builds and full suites on macOS arm64 / LLVM 22 / native (1979 successes, 0 unexpected failures) and x86-64 Linux / LLVM 18 / bytecode (1979, 0 unexpected failures, 13 consecutive runs). `SLAD-SNAPSHOT` and `SLAD-EXECUTABLE` pass on both, so the resolution is exercised rather than merely reviewed — and the `__x86_64__` objcopy branch is covered by the Linux run.

Nothing to change in this PR; just flagging it so whoever merges it does not resolve the conflict by taking one side.

PR #1777, opened 2026-05-27
Trampolines

Bike:
Adds a facility for giving bytecodes native "trampolines" so that they are visible by name in external backtraces (e.g. perf, gdb). Supersedes #1765. drmeister has done the work here, just filing a PR so I can review it.


include/clasp/core/clasp_gmpxx.h
```diff
@@ -28,6 +28,9 @@ THE SOFTWARE.
 
 /* Define a C++ GMP wrapper */
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-literal-operator"
```
Bike, line 36: comment as to what this is for would be appreciated

include/clasp/core/sampling_profiler.h
```diff
@@ -0,0 +1,97 @@
...
+ * running thread. The handler walks the frame-pointer chain via the
+ * ucontext registers and appends a sample (timestamp, thread id, depth,
+ * optional bytecode-VM pc, variable-length PC array) to a per-process
+ * bump-allocated ring.
+ *
+ * Separate from src/core/profiler.cc's RangePush/RangePop instrumentation.
+ * That profiler measures user-annotated regions; this one periodically
+ * snapshots whatever code is running.
+ *
+ * See Phase 4 / Phase 5 for post-mortem symbolication and flame-graph
```
Bike, line 14: What phases are this referring to? (Probably phases of a Claude plan, but that's not apparent from the code, so.)

include/clasp/core/sampling_profiler.h
```diff
@@ -0,0 +1,97 @@
...
+// Return one entry per recorded sample. Each inner vector holds the
+// symbolicated frame names for that sample, outermost-first (index 0
+// is the root, last index is the leaf). Prints a warning and returns
+// an empty vector if the profiler is still running.
+std::vector<SymbolicatedSample> sampling_profiler_symbolicated_samples();
+
+// Populate the calling thread's stack bounds for later frame-walking.
+// Must be called from a non-signal context. sampling_profiler_start
+// calls this automatically for the calling thread; other threads that
+// should be fully profiled need to call ext:profile-register-thread
```
Bike: I don't think they do, since mpPackage.h does for all threads?

src/core/trampoline/trampoline.cc
```diff
@@ -33,6 +40,15 @@ return_type WRAPPER_NAME(uint64_t pc, void *closure, uint64_t nargs, void **args
...
+  return g_bytecode_call_ptr(pc, closure, nargs, args);
 }
+
+// End marker placed directly after the trampoline body so we can compute
+// the trampoline's compiled size at runtime: size = &end_marker - &wrapper.
+// __attribute__((used)) prevents the optimizer from dropping it; the asm
+// name participates in the same "wrapper:name" -> unique-name substitution
+// (the suffix "_end" survives unchanged), so each trampoline gets its own
+// matching end marker symbol.
+__attribute__((used, noinline)) void WRAPPER_END_MARKER() asm("wrapper:name_end");
```
Bike: since we're hardcoding the instructions, none of this should be necessary, and i don't see it being used

src/core/trampoline/trampoline.cc
```diff
@@ -21,7 +21,14 @@ void *CLASP_LITERALS(trampoline)[0];
 // Use asm and nodebug to add declaration for the LLVM intrinsic.
 __attribute__((nodebug)) void LLVM_EXPERIMENTAL_STACKMAP(uint64_t type, uint32_t dummy, ...) asm("llvm.experimental.stackmap");
 
-return_type bytecode_call(uint64_t pc, void *closure, uint64_t nargs, void **args);
+// Indirect through a global function pointer rather than calling bytecode_call
+// directly. Each compiled trampoline's call topology is now identical
```
Bike: "now identical" - claude seems prone to putting progress reports in comments, but they'd be more appropriate in commits

src/core/bytecode.cc
```diff
@@ -1581,6 +1581,14 @@ gctools::return_type bytecode_call(unsigned char* pc, core::T_O* lcc_closure, si
   }
 }
 
+// Legacy-path hook retained for src/core/trampoline/trampoline.cc which is
+// still compiled into libclasp (its IR is also embedded as global_trampoline
+// for target-datalayout extraction). The arena-mode trampolines no longer
+// indirect through this pointer — they call bytecode_call via an absolute
+// address baked into the trampoline template by LLVM. Will be removed when
+// the legacy LLVM trampoline path is deleted.
```
Bike: okay, then let's just do that now?

src/gctools/snapshotSaveLoad.cc
```diff
@@ -1150,6 +1168,11 @@ struct copy_objects_t {
       if (header->_badge_stamp_wtag_mtag._value == DO_SHIFT_STAMP(gctools::STAMPWTAG_llvmo__ObjectFile_O)) {
         llvmo::ObjectFile_O* objectFile = (llvmo::ObjectFile_O*)clientStart;
         llvmo::ObjectFile_O* code = (llvmo::ObjectFile_O*)clientStart;
+        // Defense-in-depth: transient arena-init scaffolding must never make
```
Bike: what the heck does "defense in depth" mean here? If an earlier check should catch this where is that?

src/gctools/snapshotSaveLoad.cc
```diff
@@ -2362,8 +2414,10 @@ void snapshot_load(void* maybeStartOfSnapshot, void* maybeEndOfSnapshot, const s
         // Link all the code objects
         MaybeTimeStartup time5("Object file linking");
         using TP = thread_pool<ThreadManager>;
-#if 0
+#if 1
```
Bike, line 2434: I think we were already down to one thread through a lock. That's ok but then we should remove the lock code if we have this.

include/clasp/llvmo/trampoline_x86_64.h
```diff
@@ -0,0 +1,152 @@
+// Hardcoded x86_64 trampoline templates.
+// One shared CIE, separate code+FDE per kind (bytecode vs GF).
+//
+// At init time: copy the code array, patch the 8-byte target address at the
+// known offset, then hand code+CIE+FDE to the arena's install_template().
+// No LLVM JIT compilation needed.
```
Bike: another unnecessary response comment

include/clasp/llvmo/trampoline_x86_64.h
```diff
@@ -0,0 +1,152 @@
...
+#if defined(__x86_64__)
+
+namespace llvmo {
+namespace trampoline_x86_64 {
+
+// =========================================================================
+// Bytecode trampoline (26 bytes)
+// Signature: (i64 pc, ptr closure, i64 nargs, ptr args) -> {ptr, i64}
+// Passes all args through; saves pc (%rdi) on stack for debugger visibility.
+// The sub $16 (not push) keeps 16-byte stack alignment for the callee.
```
Bike: how does subtracting 16 affect alignment to 16 bytes? if it wasn't already aligned the sub wouldn't make it more aligned.

src/core/backtrace.cc
```diff
@@ -433,33 +434,84 @@ static DebuggerFrame_sp make_lisp_frame(size_t frameIndex, void* absolute_ip, co
...
   T_sp spi = bytecode_spi_for_pc(fun->code(), bpc);
-  // Grab arguments.
-  T_sp tnargs((gctools::Tagged)(*(bfp - BYTECODE_FRAME_NARGS_OFFSET)));
-  size_t nargs = tnargs.unsafe_fixnum();
-  T_O** argptr = (T_O**)*(bfp - BYTECODE_FRAME_ARGS_OFFSET);
+  // Recover arguments from the parent trampoline frame's saved-arg area.
+  // The VM stack also has nargs/args at fp - BYTECODE_FRAME_NARGS_OFFSET /
+  // ARGS_OFFSET (pushed by bytecode_call for the chain walk), but the
+  // trampoline path also yields the closure and works uniformly for every
+  // bytecode_call C frame.
```
Bike: we should use the old method at least as a backup for when trampolines aren't available.

src/core/backtrace.cc
```diff
@@ -550,26 +603,76 @@ static DebuggerFrame_sp make_cxx_frame(size_t fi, void* ip, const char* cstring,
...
-    return make_bytecode_frame(fi, bytecode_pc, bytecode_fp);
+    return make_bytecode_frame(fi, bytecode_pc, bytecode_fp, fbp);
   T_sp lname = SimpleBaseString_O::make(name);
   D(printf("%s%s:%d:%s lname %s\n", trace.spaces().c_str(), __FILE__, __LINE__, __FUNCTION__, name.c_str()););
   return DebuggerFrame_O::make(lname, Pointer_O::create(ip), nil<T_O>(), nil<T_O>(), nil<T_O>(), nil<T_O>(), false, nil<T_O>(),
                                INTERN_(kw, c_PLUS__PLUS_), false);
 }
 
+// Arena-trampoline frame: the PC lives in an mmap'd arena slot, so there's
+// no ObjectFile and no DWARF. The side table gives us the trampoline's name.
```
Bike: there is DWARF or why would we have an FDE and CIE? i don't think we can get at it here, though, so the comment is just a bit wrong

src/llvmo/trampolineWork.cc
```diff
@@ -0,0 +1,363 @@
...
+#include <clasp/core/foundation.h>
+#include <clasp/core/object.h>
+#include <clasp/core/pointer.h>
+#include <clasp/core/bytecode.h>             // bytecode_call (call target of every trampoline)
+#include <clasp/core/wrappers.h>             // Values() multi-value ctor
+#include <clasp/core/lisp.h>                 // _lisp root, used by the post-load regen pass
+#include <clasp/core/compiler.h>             // comp::_sym_STARthread_safe_contextSTAR
+#include <clasp/core/funcallableInstance.h>
+#include <clasp/llvmo/trampoline_arena.h>    // arena-mode trampolines
+#include <clasp/llvmo/trampoline_x86_64.h>   // hardcoded x86_64 trampoline templates
```
Bike, line 25: probably should be conditional by architecture?

src/llvmo/trampoline_arena.cc
```diff
@@ -0,0 +1,495 @@
...
+ * padded to a 16-byte stride. The caller supplies pre-composed tramp/CIE/FDE
+ * byte arrays, already patched so that:
+ *   - FDE's PC begin uses pcrel|sdata4 encoding whose value equals the
+ *     compile-time-constant distance back to the code (= -(code_size +
+ *     cie_size + 8)).
+ *   - FDE's CIE pointer equals the compile-time-constant distance back to
+ *     the preceding CIE (= cie_size + 4).
+ *   - FDE's PC range = code_size.
+ * Because every slot has the same layout, the bytes are identical across
+ * slots and a single memcpy is all that's required.
```
Bike: phrased misleadingly - each copy needs to be patched with some addresses, so in the end they will have different bytes. And while there's only one explicit memcpy call, the template is copied into a `std::vector` first, which probably does the like of memcpy.

src/llvmo/trampoline_arena.cc
```diff
@@ -0,0 +1,495 @@
...
+static void gdb_jit_register(const char* elf_buf, size_t elf_size);
+
+class TrampolineArenaInstance {
+public:
+  TrampolineArenaInstance(const char* label) : _label(label) {}
+
+  bool install_template(const uint8_t* tramp_bytes, size_t tramp_size,
+                        const uint8_t* cie_bytes,   size_t cie_len,
+                        const uint8_t* fde_bytes,   size_t fde_len) {
+    std::lock_guard<std::mutex> g(_init_lock);
```
Bike, line 241: Checking if `_initialized` is true first (presumably with acquire) should be faster than grabbing a lock every time we compile anything.

src/lisp/kernel/lsp/flamegraph.lisp
```diff
@@ -0,0 +1,1142 @@
...
+        for pos = (position sep s :start start)
+        if pos
+          collect (subseq s start pos) into parts
+          and do (setf start (1+ pos))
+        else
+          return (nconc parts (list (subseq s start len)))))
+
+(defun join (list sep)
+  (with-output-to-string (out)
+    (loop for cell on list
```
Bike: `(loop for (item . rest) on list do (write-string item out) when rest do (write-char sep out))`

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,866 @@
...
+// Read CLOCK_MONOTONIC in nanoseconds. clock_gettime is async-signal-safe
+// on Linux and macOS for CLOCK_MONOTONIC.
+static inline uint64_t now_ns_signal_safe() {
+  struct timespec ts;
+  clock_gettime(CLOCK_MONOTONIC, &ts);
+  return (uint64_t)ts.tv_sec * 1000000000ull + (uint64_t)ts.tv_nsec;
+}
+
+// Read the interrupted instruction pointer out of the context structure.
+// x86_64 only for now — portable to arm64 when we need it.
```
Bike, line 217: so... not portable?

src/core/sampling_profiler.cc
```diff
@@ -0,0 +1,866 @@
...
+
+struct ThreadStackBounds {
+  uintptr_t lo;
+  uintptr_t hi;
+  bool populated;
+};
+
+thread_local ThreadStackBounds t_stack_bounds{0, 0, false};
+
+static void populate_stack_bounds_for_this_thread() {
```
Bike, line 273: I already do almost exactly this for scanning in the mmtk branch. so I guess that'll have to get merged eventually

Bike (commented):
mostly seems ok. some of the comments are false or misleading. I didn't read over the sampling profiler or flame graphs in great detail, but they're optional, so I'm not too concerned about those.

of course, tests are failing, so that's no good

include/clasp/core/clasp_gmpxx.h
drmeister: I added this comment
https://github.com/clasp-developers/clasp/blob/trampolines/include/clasp/core/clasp_gmpxx.h#L31

include/clasp/core/sampling_profiler.h
drmeister: Phase 4 and Phase 5 are mentioned here:
https://github.com/clasp-developers/clasp/blob/trampolines/src/core/sampling_profiler.cc#L557
They are symbolication and collapsed-stacks aggregation.

include/clasp/core/sampling_profiler.h
drmeister: I'm not sure what you are asking.

I added sampling_profiler_register_current_thread so the sampling_profiler has easy access to the thread stacks that we want to walk.  It was easier than figuring how to feed it our existing stack bounds.  It's 40 bytes for each lisp thread.  I added calls for the main thread and all threads launched in mpPackage.cc

src/core/trampoline/trampoline.cc
drmeister: I removed it.

src/core/trampoline/trampoline.cc
drmeister: The file is removed.

src/llvmo/trampoline_arena.cc
drmeister: I updated the comment.  The trampolines use 8-byte absolute address to bytecode_vm so they don't need to be individually fixed up.

src/llvmo/trampoline_arena.cc
drmeister: I think this is only called once at startup.  Not every time we compile anything.

src/lisp/kernel/lsp/flamegraph.lisp
drmeister: Changed - thanks.

src/core/sampling_profiler.cc
drmeister: I haven't tested it on arm64 yet.

src/core/sampling_profiler.cc
drmeister: Ok.

include/clasp/core/sampling_profiler.h
Bike: I mean, the comment says that threads need to call `ext:profile-register-thread`. But they actually don't, since you set it up already in all thread launches in mpPackage.cc and for the main thread.

drmeister: I shortened this comment to remove the statement that it needs to be called.

src/core/bytecode.cc
drmeister: I removed this comment previously.

src/gctools/snapshotSaveLoad.cc
drmeister: I think it means there are multiple layered tests to avoid trampolines being written into the snapshots and if this one triggers then an outer layer is failing.  I added a warning.

include/clasp/llvmo/trampoline_x86_64.h
drmeister: I removed the "No LLVM JIT compilation needed." comment.  The rest of it is ok.

include/clasp/llvmo/trampoline_x86_64.h
drmeister: Removed line 21

src/core/backtrace.cc
drmeister: Removed mention of "no ObjectFile and no DWARF".

src/core/backtrace.cc
drmeister: I added code to use the old method as a backup.

src/gctools/snapshotSaveLoad.cc
drmeister: I removed the lock code.  I'm leaving in the #if 1/#else/#endif

Bike:
So the build fails on mac because this unconditionally builds trampoline_work.cc which is full of elf specifics, including an #include elf.h, and elf.h doesn't exist on mac. should the entire file be conditionalized or what?

drmeister:
I fixed the trampoline problems on macos so it builds for me.
I'm pushing changes now.

PR #1783, opened 2026-06-01
Don't register EHFrameRegistrationPlugin on MachO (fixes #1782)

dg1sbg:
## Problem

The long-standing `clasp`/`cando` **macos-latest/native** "Run regression tests" CI failure (#1782): the `mp` regression suite reaches `PROCESS-ABORT-1` and the process `std::terminate`s instead of unwinding (`startRunStop.cc:135 unhandled unknown exception` → `Abort signal`). Ubuntu native and the bytecode image are unaffected — only the **native image on arm64-darwin**.

## Root cause

On macOS/arm64, native (cleavir-compiled) code unwinds via **compact-unwind** (`__TEXT,__unwind_info`), not DWARF `.eh_frame` — clasp's native MachO objects carry **no** `__eh_frame`. `ClaspJIT_O` adds `orc::EHFrameRegistrationPlugin` to its `ObjectLinkingLayer`. But that plugin's MachO path inserts a *front* PrePrune pass that **removes the `__LD,__compact_unwind` section**:

```cpp
// llvm/lib/ExecutionEngine/Orc/EHFrameRegistrationPlugin.cpp
if (LG.getTargetTriple().isOSBinFormatMachO())
  PassConfig.PrePrunePasses.insert(PassConfig.PrePrunePasses.begin(),
    [](LinkGraph &G) {
      if (auto *CUSec = G.findSectionByName(MachOCompactUnwindSectionName))
        G.removeSection(*CUSec);
      return Error::success();
    });
```

That deletes the only unwind info native code has, **before** JITLink's `CompactUnwindManager` can synthesize `__TEXT,__unwind_info`. libunwind is then left with no description for native PCs, so a C++ exception thrown below a native frame (`mp::abort_process` → `throw AbortProcess()`) and caught above it (`Process_O::runInner`) cannot find its handler → `std::terminate` → abort.

(For the record: the `__compact_unwind` records are valid self-sufficient `UNWIND_ARM64_MODE_FRAME` entries; the original "objects lack `__eh_frame`" framing was a red herring — they correctly lack it, and `0x04000000` is `MODE_FRAME` on arm64, not `MODE_DWARF`.)

## Fix

[...]

Issue #1782, opened 2026-05-31
macOS arm64: EHFrameRegistrationPlugin strips __compact_unwind → C++ exceptions can't unwind across native frames (mp PROCESS-ABORT aborts CI)


dg1sbg:
## Summary

On **macOS arm64**, a C++ exception thrown below a native (cleavir-compiled) frame and caught above it cannot find its handler — the process `std::terminate()`s → `abort`. This is the long-standing `cando`/`clasp` **`macos-latest`/native** "Run regression tests" CI failure (`mp` `PROCESS-ABORT`). The Ubuntu native job and the bytecode image are fine — only the **native image on arm64-darwin** is affected.

```
../src/gctools/startRunStop.cc:135 There was an unhandled unknown exception in process - do something about it.
Terminating Clasp due to interrupt: Abort signal.
```

## Root cause (corrected — see note at bottom)

On macOS/arm64, native code unwinds via **compact-unwind** (`__TEXT,__unwind_info`), not DWARF `.eh_frame` — clasp's native MachO objects carry **no** `__eh_frame`.

`ClaspJIT_O` (`src/llvmo/runtimeJit.cc`) adds `orc::EHFrameRegistrationPlugin` to its `ObjectLinkingLayer`. That plugin's MachO path inserts a *front* PrePrune pass that **removes the `__LD,__compact_unwind` section** before anything else runs:

```cpp
// llvm/lib/ExecutionEngine/Orc/EHFrameRegistrationPlugin.cpp
if (LG.getTargetTriple().isOSBinFormatMachO())
  PassConfig.PrePrunePasses.insert(PassConfig.PrePrunePasses.begin(),
    [](LinkGraph &G) {
      if (auto *CUSec = G.findSectionByName(MachOCompactUnwindSectionName))
        G.removeSection(*CUSec);
      return Error::success();
    });
```

This deletes the only unwind info native code has, **before** JITLink's `CompactUnwindManager` (`MachO_arm64.cpp`) can synthesize `__TEXT,__unwind_info` from it. libunwind is then left with no description for native PCs, so the C++ throw in `mp::abort_process` (`throw AbortProcess()`, `mpPackage.cc`) cannot unwind up to its `catch (AbortProcess&)` in `mp::Process_O::runInner` → `std::terminate`.

[...]

PR #1785, opened 2026-06-01
Fix #1784: route mp:abort-process/exit-process via SJLJ on macOS arm64

### LINKED ISSUE #1784: macOS arm64: mp:process-cancel aborts — C++ exception can't unwind across JIT'd native cleavir frames in the interrupt dispatch
## Summary

On **macOS arm64 native** (`boehmprecise`), the `interrupt` regression suite aborts at its very first test, `CANCELLATION-INTERRUPT`, with:

```
../src/gctools/startRunStop.cc:135 There was an unhandled unknown exception in process - do something about it.
Terminating Clasp due to interrupt: Abort signal.
```

This is a **separate** problem from #1782 (synchronous `mp:abort-process`, fixed in #1783). It was previously masked: the `mp` suite aborted first at `PROCESS-ABORT`; once #1783 lets `mp` pass, CI reaches `interrupt` and hits this.

## Minimal reproduction

```lisp
(let ((lock (mp:make-lock)) proc)
  (mp:with-lock (lock)
    (setf proc (mp:process-run-function 'p
                 (lambda () (mp:with-lock (lock) (core:check-pending-interrupts)))))
    (sleep 0.2)
    (mp:process-cancel proc))
  (ignore-errors (mp:process-join proc)))
```
→ `std::terminate` / abort on macOS arm64 native.

## Root cause

The cancellation is delivered **asynchronously via a signal (SIGCONT)**. The signal handler runs the queued interrupt *in signal context*, which calls `mp:abort-process` → `throw AbortProcess()`. That C++ exception must then unwind **across the signal trampoline** (`_sigtramp`) to reach `catch (AbortProcess&)` in `mp::Process_O::runInner` — and that crossing fails, so `__cxa_throw` finds no handler → `std::terminate`.

lldb backtrace at the abort (bottom of the 71-frame stack):
```
#7  mp::mp__PERCENTabort_process            mpPackage.cc:556   (throw AbortProcess())
... deeply nested bytecode_vm + native cleavir frames ...
#66 gctools::handle_queued_interrupt        interrupt.cc:138
#67 gctools::handle_queued_interrupts       interrupt.cc:177
#68 gctools::handle_SIGCONT(sig=19)         interrupt.cc:212
#69 libsystem_platform.dylib`_sigtramp                         <- signal trampoline
```

Importantly this is **not** the #1782 cause (missing/stripped compact-unwind): with #1783 applied, libunwind's dynamic registry *does* find all the native cleavir frames (`UnwindInfoManager::findSectionsImpl` returns "found" for each). The failure is specifically **throwing a C++ exception out of a signal handler and unwinding across `_sigtramp`**, which is unreliable on macOS arm64.

## Suggested direction

Process the queued interrupt at a **synchronous safe point** rather than throwing from the signal handler — e.g. have `handle_SIGCONT` only set the pending flag / wake the thread, and let the synchronous `core:check-pending-interrupts` raise the cancellation (so the throw originates from normal code, not signal context). Alternatively, unwind the cancellation via clasp's SJLJ mechanism instead of a raw C++ throw. (Linux is unaffected because ELF `.eh_frame` + its signal-frame handling differ.)

## Environment

Apple silicon, macOS 15, Homebrew LLVM 22, `boehmprecise` native image. Same `clasp`/`cando` `macos-latest`/native CI job as #1782.

dg1sbg:
## Problem (#1784)

On macOS arm64 (native image), the `interrupt` regression suite aborts the whole process at its first test, `CANCELLATION-INTERRUPT` (`mp:process-cancel`), with `startRunStop.cc:135 ... unhandled unknown exception` → `Abort signal`.

## Root cause

A deep C++ `AbortProcess` throw that originates **inside clasp's interrupt-dispatch machinery** (`check-pending-interrupts` → `handle_queued_interrupt` → `funcall(signal-interrupt)` → … → `mp:abort-process` → `throw AbortProcess()`) cannot be unwound by macOS-arm64 libunwind across the JIT'd native cleavir frames in that path. The `__unwind_info`-driven EH unwind **derails** before reaching `catch (AbortProcess&)` in `Process_O::runInner`, so `__cxa_throw` falls through to `std::terminate`.

Evidence (all deterministic):
- A **signal-free** repro — a process that self-interrupts (`mp:interrupt-process (mp:current-process) #'mp:abort-process`) and drains via `core:check-pending-interrupts` — terminates identically, with **no `_sigtramp` on the stack**. So this is *not* the "throw across the signal trampoline" hazard; it's the C++ EH unwinder failing on the JIT frames.
- The **same dispatch path** raising a Lisp `error` under `handler-case` is caught fine, because clasp's Lisp unwinder (`_longjmp`, no unwind tables) crosses those frames where the C++ EH unwinder does not.
- A personality (`__gxx_personality_v0`) walk shows phase-1 reaching the bytecode VM frames but never `runInner`'s catch; `findSectionsImpl` stops being consulted at the offending JIT frame.

This is distinct from #1782/#1783 (which fixes *shallow* C++ throws across native frames by registering compact-unwind); the deep interrupt-dispatch path still derails, and this fix sidesteps it rather than relying on the EH unwinder at all.

## Fix

Route `mp:abort-process` and `mp:exit-process` through clasp's own **SJLJ** non-local exit (`core::sjlj_throw` to the process object as the catch tag, with the matching catch established via `call_with_catch` in `Process_O::runInner`) instead of `throw AbortProcess()` / `throw ExitProcess()`. `_longjmp` uses no unwind tables, so it is immune to the derail — and it is exactly how `cl:throw`/`handler-case` already unwind across these frames.

- Gated to `_TARGET_OS_DARWIN`. **Non-Darwin keeps the C++ throw path byte-for-byte unchanged.**
- A `completed_normally` flag set as the last statement of the catch lambda distinguishes a normal return from a non-local exit (whose `_Aborted`/`_AbortCondition` or `_ReturnValuesList` state is set before unwinding).
- One file, three functions. The throw sites (`mpPackage.cc`) and the only catcher (`runInner`) are all in this file.

## Verification (macOS arm64, native boehmprecise image)

- `TEST_SUITES=mp,interrupt` → **52 successes, 0 failures**. `CANCELLATION-INTERRUPT` now passes, as do `SLEEP/LOCK/INPUT-INTERRUPTIBLE` and `UNWIND-PROTECT.INTERRUPT.1/2`, and all `PROCESS-ABORT-*`/`PROCESS-EXIT`/`ATOMIC-*`.
- Broader sweep (`mp,interrupt,unwind,conditions,control01,ehkiller,clos,update-instance-abort`) → **135 successes, 0 failures**, no regressions.

Companion to #1783 (#1782); together they take the macOS-arm64 native CI green through the `interrupt` suite.

### DIFF
--- a/src/core/mpPackage.cc
+++ b/src/core/mpPackage.cc
@@ -157,6 +157,30 @@ void Process_O::runInner(core::List_sp bindings) {
   } else {
     updatePhase(Running);
     core::T_mv result_mv;
+#ifdef _TARGET_OS_DARWIN
+    // issue #1784: On macOS arm64 a C++ exception cannot be unwound across the
+    // JIT'd native cleavir frames in clasp's interrupt-dispatch path — libunwind
+    // derails before reaching the catch below, so a raw `throw AbortProcess()`
+    // from a delivered interrupt (e.g. mp:process-cancel) reaches std::terminate
+    // and aborts the whole process. clasp's own SJLJ non-local exit DOES cross
+    // those frames (it's how cl:throw/handler-case already unwind here). So on
+    // Darwin mp:abort-process / mp:exit-process do sjlj_throw to this process
+    // object as the catch tag instead of throwing; we establish the matching
+    // catch here. The thrower sets _Aborted/_AbortCondition (abort) or
+    // _ReturnValuesList (exit) before unwinding; completed_normally stays false
+    // unless apply0 returns without a non-local exit.
+    bool completed_normally = false;
+    result_mv = core::call_with_catch(this->asSmartPtr(), [&]() -> core::T_mv {
+      core::T_mv r = core::core__apply0(core::coerce::calledFunctionDesignator(_Function), _Arguments);
+      completed_normally = true;
+      return r;
+    });
+    if (!completed_normally) {
+      // Reached via mp:abort-process or mp:exit-process; their state is set.
+      updatePhase(Exited);
+      return;
+    }
+#else
     try {
       result_mv = core::core__apply0(core::coerce::calledFunctionDesignator(_Function), _Arguments);
     } catch (ExitProcess& e) {
@@ -172,6 +196,7 @@ void Process_O::runInner(core::List_sp bindings) {
       updatePhase(Exited);
       return;
     }
+#endif
     updatePhase(Exited);
     ql::list return_values;
     int nv = result_mv.number_of_values();
@@ -540,7 +565,13 @@ DOCGROUP(clasp)dx");
 CL_DEFUN void mp__exit_process(core::List_sp values) {
   Process_sp this_process = gc::As<Process_sp>(_sym_STARcurrent_processSTAR->symbolValue());
   this_process->_ReturnValuesList = values;
+#ifdef _TARGET_OS_DARWIN
+  // issue #1784: SJLJ non-local exit instead of a C++ throw that can't unwind
+  // across JIT cleavir frames on macOS arm64. See Process_O::runInner.
+  core::sjlj_throw(this_process);
+#else
   throw ExitProcess();
+#endif
 };
 
 // See abort-process in mp.lisp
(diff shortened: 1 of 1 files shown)


Bike:
This PR's more of a problem than the others.
While the explanation of the bug seems plausible, the correct fix is for the underlying problem - making it possible to throw through JITted frames on Macs. As I recall they recently fixed up LLVM ORC with the mac unwinding format, so we can probably do that. I guess that's #1783?
That said, it does make sense to implement abort-process through Clasp unwinding rather than C++ exceptions. It's faster and less prone to `std::terminate`. Neither of those reasons is OS-specific so there's no reason to gate on `_TARGET_OS_DARWIN`, and doing so is an unnecessary complication.
Doing this by throwing and catching the process object is no good. There's nothing preventing the programmer from doing `(catch mp:*current-process* ...)` and thereby redirecting `abort-process`, and they might even do that accidentally. It would be better to do block/return-from, but that would require an extra slot somewhere for the block. `catch` but with some less accessible object would be fine - say an internal `mp` symbol.

PR #1786, opened 2026-06-02
Fix Apple Silicon W^X SIGBUS in setf_jit_lookup_t (FFI callbacks)

dg1sbg:
## Problem

On Apple Silicon, JIT'd code/data lives in `MAP_JIT` memory that is write-protected (execute mode) by default. A thread must switch it to write mode (`pthread_jit_write_protect_np`) around any store, or the store faults with `SIGBUS` (`KERN_PROTECTION_FAILURE`).

`setf_jit_lookup_t` (`llvmo:jit-lookup-t` setf, `src/llvmo/llvmoPackage.cc`) stores a Lisp function pointer into a JIT-emitted global. `make-callback` / `clasp-ffi:%defcallback` reach it via `(setf (llvm-sys:jit-lookup-t dylib varname) function)`, and on arm64-darwin this store **SIGBUSes at compile time** — which is what makes the `defcallback-native` regression test (`CFFI-DEFCALLBACK`) crash during `compile-file`.

## Fix

Wrap the store in `JITDataReadWriteMaybeExecute()` / `JITDataReadExecute()`, matching the existing W^X guards on the other JIT-literal write sites (`core__literals_vset` in compiler.cc, `op_setf_literals` / `attr_clasp_module_native` in loadltv.cc). This is the one remaining unguarded JIT-data store, unique to the FFI-callback path.

## Verification

macOS arm64, native `boehmprecise` image: `CFFI-DEFCALLBACK` now **passes** (previously a Bus error during compile-file). Full regression suite: no regressions.

### DIFF
--- a/src/llvmo/llvmoPackage.cc
+++ b/src/llvmo/llvmoPackage.cc
@@ -55,6 +55,7 @@ THE SOFTWARE.
 #include <clasp/llvmo/debugInfoExpose.h>
 #include <clasp/llvmo/intrinsics.h>
 #include <clasp/llvmo/claspLinkPass.h>
+#include <clasp/llvmo/code.h> // JITDataReadWriteMaybeExecute / JITDataReadExecute (W^X)
 #include <clasp/core/bytecode.h>
 #include <clasp/core/instance.h>
 #include <clasp/core/funcallableInstance.h>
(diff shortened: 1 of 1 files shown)


Bike:
merged in #1794

PR #1788, opened 2026-06-02
Fix snapshot load on macOS arm64 (W^X + relocation decode + exec-link quoting)

dg1sbg:
Makes snapshot **load** work on macOS arm64. With #1787 (which fixes the *save* unwinding), `SLAD-SNAPSHOT` and `SLAD-EXECUTABLE` both pass; full regression suite 1957/11 (the 11 are pre-existing, orthogonal).

Three independent fixes, all in `snapshotSaveLoad.cc`:

## 1. W^X faults during in-place fixup (commit 1)

`snapshot_load` fixes up the loaded code **in place** in MAP_JIT (rwx, W^X) memory; on Apple Silicon a bare store into execute-mode JIT memory faults with `SIGBUS`/`KERN_PROTECTION_FAILURE`. Three sites now switch to write mode via `JITDataReadWriteMaybeExecute()`/`JITDataReadExecute()` (same pattern as the existing JIT-literal write sites): the code-literals `memcpy`, and the `fixup_objects` / `fixup_internals` walks. (The *save* path fixes up a RW copy, so it never hit this.)

## 2. Relocation-decode false positive (commit 2)

The entry-point decode validated the resolved address by comparing its first byte against a saved `firstByte`, and **aborted** on mismatch (`decodeEntryPointForCompiledCode`) / warned (`fixedAddress`). But a function's first instruction is frequently a relocatable instruction (e.g. **ADRP**), whose encoded page-offset immediate legitimately differs between the save-time and load-time JIT (different load addresses). The offset/symbol-based decode is correct — verified: with the check made non-fatal the snapshot loads, runs, and returns the right value — so the `firstByte` equality check is a false positive on every relocated first instruction. Dropped at both sites.

## 3. Executable-link path quoting (commit 2)

`save-lisp-and-die :executable` builds a `clang++` command run via `system()` (a shell). The `-Wl,-force_load,<libdir>/libiclasp.a` path was unquoted, so a build/lib dir containing a space (e.g. a Dropbox path) split the arg and the link failed (`no such file`). The output, `-sectcreate` file, and `-force_load` paths are now quoted.

## Verification (macOS arm64, native boehmprecise image)

`SLAD-SNAPSHOT` and `SLAD-EXECUTABLE` pass; full suite run from `build/` = **1957 passed / 11 failed** (the 11 known pre-existing). No regressions.

### DIFF
--- a/src/gctools/snapshotSaveLoad.cc
+++ b/src/gctools/snapshotSaveLoad.cc
@@ -2598,7 +2598,14 @@ void snapshot_load(void* maybeStartOfSnapshot, void* maybeEndOfSnapshot, const s
         //
         if (oldCodeClient->literalsSize() != 0 &&
             (newCodeDataStart <= newCodeLiteralsStart && newCodeLiteralsEnd <= newCodeDataEnd)) {
+          // The new code's literal vector lives in the CodeBlock's MAP_JIT data
+          // region, which on Apple Silicon is write-protected (execute mode) by
+          // default; switch this thread to write mode around the store or snapshot
+          // LOAD faults with SIGBUS (KERN_PROTECTION_FAILURE), same W^X hazard as the
+          // other JIT-literal write sites (compiler.cc literals_vset, loadltv, etc.).
+          llvmo::JITDataReadWriteMaybeExecute();
           memcpy((void*)newCodeLiteralsStart, (void*)oldCodeClient->literalsStart(), newCodeClient->literalsSize());
+          llvmo::JITDataReadExecute();
         }
         //
         // Now set the forwarding pointer from the oldCode object to the newCodeClient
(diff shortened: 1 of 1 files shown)


dg1sbg:
Added a 4th commit (`save-lisp-and-die :executable` absolute `-L`): the executable link command carried only the **relative** `-Lboehmprecise/lib` from `BUILD_LINKFLAGS`, so `-lclasp` resolved only when run with `CWD=build/`; from any other directory the link failed with "library not found for -lclasp". Now adds an absolute `-L"<_LibDir>"` (matching the Linux branch). Verified: a standalone executable created and run from `/tmp` returns the right value, and `SLAD-EXECUTABLE` passes from the repo root, not just `build/`.

Bike:
1, the snapshot load change, appears to be duplicating the similar change in #1768.

PR #1789, opened 2026-06-03
Define cmp:*compile-native* flag for controlling COMPILE behavior

Bike:
frgo/@dg1sbg wants this for constrained environments - see #1769. And it's good to have anyway, and cleaner than binding `*btb-compile-hook*`.

`*btb-compile-hook*` is removed. It was a vestige of the old build process anyway.


Bike:
cando failure is unrelated (depends on `core:fsync` which is not yet in main)

PR #1790, opened 2026-06-04
A few additions to main

drmeister:
I added core__fsync and core__fsync-directory to try and speed up Network File System updates on the clusters I work on.

I added the ext:float-nan-string function to return different kinds of NAN values.


Bike:
https://github.com/clasp-developers/clasp/actions/runs/26923198904/job/79427805072?pr=1790#step:16:52 This is the only potentially relevant CI problem - any idea what's up with it?

Bike:
It's a problem in Cando - the tests were changed to hardcode the quicklisp location https://github.com/cando-developers/cando/commit/4a553d0099fd081e94085d50597e99257add945a#diff-92eb23c031e0066f14dfc0cb5689ecce968cb233d3ada65e69b31cb617a874c7R15
but that's irrelevant to this PR, so merge time

PR #1792, opened 2026-06-10
Make foreign-alloc memory permanent until explicit free

dg1sbg:
## Problem

CFFI struct-by-value calls (via cffi-libffi) crash with memory corruption after a few hundred calls on Clasp — SEGV inside `ffi_call` reading a freed `ffi_type*`.

## Root cause

`%allocate-foreign-data` (fli.cc) creates the `ForeignData_O` wrapper with `DeleteOnDtor`, so the malloc'd block is `free()`d by the wrapper's GC finalizer. That ties the lifetime of the *foreign memory* to the GC-reachability of the *Lisp wrapper*.

CFFI code routinely stores the raw address inside other foreign (GC-unscanned) memory and drops the wrapper. cffi-libffi's `make-libffi-cif` is the concrete case: it caches the cif but drops the wrappers for the `arg_types` array and the per-struct `ffi_type` descriptors, whose raw pointers live inside the cif's malloc'd block. The next GC collects those wrappers, the finalizers free the still-referenced blocks, and the following call is a use-after-free.

Every other CFFI backend's `foreign-alloc` is plain malloc — permanent until `foreign-free`. Clasp's violated that contract.

Proven by pinning every `%foreign-alloc` wrapper (200k calls, zero crashes) and by forcing a GC each iteration (crash moves to iteration 0).

## Fix

Allocate with `core::None` instead of `core::DeleteOnDtor` in `PERCENTallocate_foreign_data` and `allocate_foreign_data`, so `foreign-alloc` memory persists until an explicit `%foreign-free` — matching every other CFFI backend. `with-foreign-object`'s allocator keeps `DeleteOnDtor` since it always frees explicitly on scope exit.

Blast radius: the only other in-tree `%foreign-alloc` caller (`defcallback-native.lisp`) already frees explicitly.

## Test

`src/tests/fli/cffi-fsbv-stress.lisp`: correctness across scalar / struct-return / nested-struct by-value calls plus a 300k-call stress loop. Verified on macOS arm64 (boehm): SEGVs with the fix reverted, passes with it applied.

Note: full struct-by-value support also needs a one-line fix in CFFI's `cffi-clasp.lisp` (`:default` library marker not mapped to `:rtld-default`); that goes upstream to cffi separately.

### DIFF
--- a/src/core/fli.cc
+++ b/src/core/fli.cc
@@ -460,7 +460,8 @@ void ForeignData_O::PERCENTfree_foreign_object(void) { this->free_(); }
 ForeignData_sp PERCENTallocate_foreign_data(core::Integer_sp size) {
   size_t _size = unbox_fixnum(size);
   auto self = gctools::GC<ForeignData_O>::allocate();
-  self->allocate(kw::_sym_clasp_foreign_data_kind_data, core::DeleteOnDtor, _size);
+  // foreign-alloc memory is permanent until explicit %foreign-free (CFFI/malloc contract); no GC finalizer.
+  self->allocate(kw::_sym_clasp_foreign_data_kind_data, core::None, _size);
   return self;
 }
 
@@ -468,7 +469,7 @@ ForeignData_sp PERCENTallocate_foreign_data(core::Integer_sp size) {
 // ---------------------------------------------------------------------------
 ForeignData_sp allocate_foreign_data(uint64_t size) {
   auto self = gctools::GC<ForeignData_O>::allocate();
-  self->allocate(kw::_sym_clasp_foreign_data_kind_data, core::DeleteOnDtor, size);
+  self->allocate(kw::_sym_clasp_foreign_data_kind_data, core::None, size);
   return self;
 }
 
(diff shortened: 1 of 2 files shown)


dg1sbg:
CI note: the `clasp/macos-latest/native` and `cando/ubuntu-latest/native` failures are pre-existing on `main` — the latest main run at this PR's base commit (https://github.com/clasp-developers/clasp/actions/runs/27157557544) fails the exact same steps ("Run regression tests" / "Run Cando regression tests"). Both Ubuntu clasp jobs (bytecode + native) pass on this PR.

Locally (macOS arm64, boehmprecise, base = current main + the W^X fixes from #1781/#1786/#1788): full regression suite 1963 passed / 4 failed, all four known-preexisting (SBCL-CROSS-COMPILE-4, INCLUDE-LEVEL-2B/3, TYPES-CLASSES-10), zero new failures; plus the new cffi-fsbv stress test passes (300k mixed struct-by-value calls, no crash).

Bike:
We don't really use the tests/ directory and should probably delete it, but no big

PR #1812, opened 2026-07-31
Fix three CLOS guards whose read-time class literals never match (eql-gf memoization, static slot-value, validate-superclass)

### LINKED ISSUE #1811: EQL-specialized generic functions never memoize: every call is a full dispatch miss
## Summary

Any generic function with an `eql` specializer takes a **full dispatch miss on every call**. Its call history stays permanently empty, so `compute-applicable-methods`, `compute-effective-method` and a fresh `effective-method-function` run on every single call.

Class-specialized generic functions are unaffected.

Version: `clasp-boehm-3.0.1-73-g2cf5bb5e4`, macOS arm64. Reproduces identically on `boehm` and `boehmprecise`.

## Reproducer

```lisp
(defun bytes-per (thunk n)
  (funcall thunk)
  (let ((start (gctools:bytes-allocated)))
    (dotimes (i n) (funcall thunk))
    (/ (float (- (gctools:bytes-allocated) start)) n)))

(defgeneric egf (k))
(defmethod egf ((k (eql :a))) 1)
(defmethod egf ((k (eql :b))) 2)

(bytes-per (lambda () (egf :a)) 500)                      ; => 6784.0
(length (clos::generic-function-call-history #'egf))      ; => 0   <-- never memoized
```

A class-specialized gf for comparison gives `0.0` B/call and a call history of length 1.

## Impact

`clasp-ffi:%mem-ref` / `%mem-set` define one `eql` method per foreign type (`fli.lisp`, `generate-methods`), so **every CFFI foreign memory access** hits this:

```
clasp-ffi:%mem-ref   19448 B/call, ~321 us/call, call history 0
clasp-ffi:%mem-set   20240 B/call
```

Cost scales with method count (~6300 B + ~288 B/method) and is identical for the first or last key, since it is not a search.

## Root cause

`src/lisp/kernel/clos/miss.lisp`, in `miss-info`:

```lisp
((eq (class-of generic-function)
     #.(find-class 'standard-generic-function))
 (memoize-eql-specialized generic-function method-combination
                          call-history argument-classes))
```

In the built image this comparison is always false, so `miss-info` falls through to the `(t nil)` clause and never calls `memoize-eql-specialized`. Nothing is added to the call history, so the next call misses again. Note `updatedp` is therefore always NIL and `force-discriminator` never runs — the discriminating function is not being rebuilt, the raw miss path is simply taken every time.

Replacing the read-time literal with a runtime `(find-class 'standard-generic-function)` fixes it completely: 0 B/call, and `%mem-ref` drops to 137 ns with a 27-entry call history.

## Why this is filed rather than just patched: the literal is *not* stale

Everything checkable says the comparison should succeed:

- The module literal, the live class, and `(class-of gf)` are all `eq` to one another:
  `(eq (class-of #'egf) (aref (core:bytecode-module/literals (core:bytecode-simple-fun/code #'clos::miss-info)) 9))` => `T`, and both are `eq` to `(find-class 'standard-generic-function)`.
- The bytecode decodes correctly. Raw bytes around the test: `60 8` (`called-fdefinition` EQ), `60 1` (CLASS-OF), `2 0` (`closure 0` = the gf), `4 1`, `1 9` (`const` literal 9 = the class), `4 2`, `23 12` (`jump-if-8 +12`). Offset 12 from the opcode lands exactly on the `called-fdefinition` of `MEMOIZE-EQL-SPECIALIZED`.
- Every input to the `cond` is correct. Wrapping `clos::outcome` shows, inside the compiled function's own extent: `closure 0` is `#'egf`, `(eq (class-of gf) (find-class 'standard-generic-function))` is `T`, `ok` is NIL, `final-methods` has length 1, `call-history` is NIL.
- Yet witness counters on `memoize-eql-specialized`, `specializers-combinate` and `call-history-find-key` all stay at **zero**, while a direct call to `memoize-eql-specialized` with those exact arguments returns 3 entries and trips all three counters.
- Redefining `miss-info` at runtime from **identical source** (with either the `#.` literal or a runtime `find-class`) works correctly.

So the source is right, the constant is right, the bytecode is right, and the inputs are right, but the branch is not taken in cross-clasp-compiled code. That looks like a compiler or VM issue rather than a CLOS one, which is why it is worth a look beyond the one-line workaround.

## Second occurrence of the same pattern

`src/lisp/kernel/clos/static-gfs/svuc.lisp`, `uncustomizable-slot-p`, has the same read-time-literal guard and is dead in the same way — it returns NIL for a plain `standard-class` with a `standard-effective-slot-definition`:

```lisp
(defclass plainc () ((a :initform 1)))
(let ((c (find-class 'plainc)))
  (eq (class-of c) (find-class 'standard-class)))                  ; => T
(static-gfs::uncustomizable-slot-p c slotd)                        ; => NIL, expected T
```

This silently disables the static `slot-value` / `slot-boundp` optimization for every standard class.

Not all such guards fail: `#.(find-class 't)` in `update-specializer-profile` (`generic.lisp:284`) and `#.(find-class 'standard-class)` in `std-create-slots-table` (`class.lisp:307`) both behave correctly, so it is not simply "all read-time class literals are broken".

dg1sbg:
Fixes #1811.

Three guards in CLOS compare a class against a **read-time** literal, `#.(find-class ...)`. All three evaluate false in the built image, so all three silently take their fallback path. Two cost performance; the third is a correctness bug.

## 1. `miss.lisp:266` — eql-specialized generic functions never memoize

Any generic function with an `eql` specializer takes a **full dispatch miss on every call**. Its call history stays permanently empty, so `compute-applicable-methods`, `compute-effective-method` and a fresh `effective-method-function` run every single time. Class-specialized generic functions are unaffected.

```lisp
((eq (class-of generic-function)
     #.(find-class 'standard-generic-function))
 (memoize-eql-specialized ...))
```

False in the built image, so `miss-info` falls through to `(t nil)` and never calls `memoize-eql-specialized`. Nothing is added to the call history, so the next call misses again.

This is *not* discriminating-function recomputation: `updatedp` is always NIL, so `force-discriminator` never runs. The raw miss path is simply taken every time — which is why the cost is flat across the first and last key rather than looking like a search.

| | before | after |
|---|---|---|
| 30-method eql gf, first key | 14656 B/call | **0.0 B** |
| 30-method eql gf, last key | 14656 B/call | **0.0 B** |
| eql gf call history | 0 | **31** |
| `clasp-ffi:%mem-ref` | 19448 B, 321620 ns | **0.0 B, 88 ns** |
| `clasp-ffi:%mem-set` | 20240 B | **0.0 B** |
| `%mem-ref` call history | 0 | **27** |

`clasp-ffi:%mem-ref` / `%mem-set` define one `eql` method per foreign type, so **every CFFI foreign memory access** was paying this. Cost scaled at roughly 6300 B + 288 B/method. Confirmed one layer up: compiled `(cffi:mem-ref p :float off)` goes from that to **0.00 B/call**.

## 2. `svuc.lisp:21-23` — static slot-value optimization disabled

`uncustomizable-slot-p` returned NIL for a plain `standard-class` with a `standard-effective-slot-definition`, silently disabling the static `slot-value` / `slot-boundp` optimization for **every** standard class. Both `eq` tests are true when evaluated at runtime.

## 3. `class.lisp:174-177` — `validate-superclass` rejects AMOP-legal hierarchies

This one is a correctness bug. The two clauses permitting a `standard-class` and a `funcallable-standard-class` in each other's superclass chain are both dead, so the mixed case is refused:

```lisp
(defclass plain-sc () ())
(defclass fsc-from-sc (plain-sc) () (:metaclass clos:funcallable-standard-class))
;; => Class #<STANDARD-CLASS PLAIN-SC> is not a valid superclass for
;;    #<FUNCALLABLE-STANDARD-CLASS FSC-FROM-SC>
```

Called directly, both directions returned NIL where T is required.

## Fix

Runtime `find-class` in all three. In `validate-superclass` the lookups are deferred behind the existing `(eq c1 c2)` fast path and use `errorp nil`, so the common same-metaclass case does no lookup and bootstrap cannot trip over a metaclass that is not yet registered. `cl:find-class` is a C++ `CL_DEFUN` hash lookup, not a generic function, so it is safe on the dispatch miss path.

## Verification

Six new tests, in `fastgf.lisp` and `mop.lisp`. Each fails before and passes after.

- `boehm`: **1974** successes, zero unexpected failures.
- `boehmprecise`: **1976**, zero unexpected failures.
- **Clean-bootstrap `boehmprecise`**: kernel Lisp and both images deleted and regenerated from source — no errors, **1979** successes, zero unexpected failures. This specifically checks that replacing load-time literals with runtime lookups does not destabilise bootstrap ordering, since all three sites run during bootstrap (`validate-superclass` on every `defclass`, `miss-info` on every dispatch miss, `uncustomizable-slot-p` at every kernel compile).

## Worth a second look

The change fixes the symptom; **the reason the comparisons fail is not explained**, which is why #1811 is open separately. The module literal, the live class and `(class-of x)` are all `eq` to one another; the bytecode decodes correctly; every input to the `cond` is correct — yet witness counters on the guarded functions stay at zero, and recompiling the identical source at runtime works.

Not all such guards fail: `class.lisp:184`, `class.lisp:307` and `generic.lisp:284` behave correctly. The audit in #1811 rules out per-class, per-file, `defun` vs `defmethod`, test-vs-value, and `class-of` vs `si::instance-class` as the discriminator. Whatever the real cause is, it can silently disable any future `#.(find-class ...)` guard the same way — which is exactly how the slot-value optimization sat dead without anyone noticing.

### DIFF
--- a/src/lisp/kernel/clos/class.lisp
+++ b/src/lisp/kernel/clos/class.lisp
@@ -168,13 +168,14 @@ argument was supplied for metaclass ~S." (class-of class))))))))
   supplied-superclasses)
 
 (defmethod validate-superclass ((class class) (superclass class))
+  ;; NOTE: runtime lookups (a #. literal never matches), deferred and errorp nil for bootstrap safety.
   (or (let ((c1 (class-of class))
 	    (c2 (class-of superclass)))
 	(or (eq c1 c2)
-	  (and (eq c1 #.(find-class 'standard-class))
-            (eq c2 #.(find-class 'funcallable-standard-class)))
-	  (and (eq c2 #.(find-class 'standard-class))
-            (eq c1 #.(find-class 'funcallable-standard-class)))))
+	    (let ((std (find-class 'standard-class nil))
+		  (fstd (find-class 'funcallable-standard-class nil)))
+	      (or (and (eq c1 std) (eq c2 fstd))
+		  (and (eq c2 std) (eq c1 fstd))))))
       (or (typep class 'forward-referenced-class)
           (typep superclass 'forward-referenced-class))))
 
--- a/src/lisp/kernel/clos/miss.lisp
+++ b/src/lisp/kernel/clos/miss.lisp
@@ -262,8 +262,9 @@
                       ;; another thread has already added this entry
                       nil
                       (list (cons key outcome)))))
+               ;; NOTE: must be a runtime lookup; a #. literal here never matches.
                ((eq (class-of generic-function)
-                    #.(find-class 'standard-generic-function))
+                    (find-class 'standard-generic-function))
                 (memoize-eql-specialized generic-function method-combination
                                          call-history argument-classes))
                (t
--- a/src/lisp/kernel/clos/static-gfs/svuc.lisp
+++ b/src/lisp/kernel/clos/static-gfs/svuc.lisp
@@ -16,11 +16,12 @@ TODO 2: Build a cell system like we do for make-instance.
 ;;;; NOTE: The static-foo macros may evaluate the instance form more than once.
 ;;;; (Because we know here that it's always just a variable.)
 
+;;; NOTE: must be runtime lookups; #. literals here never match.
 (defun uncustomizable-slot-p (class slotd)
   (let ((metaclass (class-of class)))
-    (and (or (eq metaclass #.(find-class 'standard-class))
-             (eq metaclass #.(find-class 'clos:funcallable-standard-class)))
-         (eq (class-of slotd) #.(find-class 'clos:standard-effective-slot-definition)))))
+    (and (or (eq metaclass (find-class 'standard-class))
+             (eq metaclass (find-class 'clos:funcallable-standard-class)))
+         (eq (class-of slotd) (find-class 'clos:standard-effective-slot-definition)))))
 
 ;;; I said "constant slotd", but slotds aren't dumpable, so to satisfy
 ;;; COMPILE-FILE what we'll actually get is (ltv-slotd class slotd),
--- a/src/lisp/regression-tests/fastgf.lisp
+++ b/src/lisp/regression-tests/fastgf.lisp
@@ -5,3 +5,12 @@
 (defmethod fgf-foo ((x symbol)) :symbol)
 (test dispatch-symbol (fgf-foo :yadda) (:symbol))
 (test-expect-error dispatch-no-applicable-method (fgf-foo 1.2) :description "This should not dispatch")
+
+(defmethod fgf-eql ((x (eql :alpha))) :alpha)
+(defmethod fgf-eql ((x (eql :beta))) :beta)
+(test dispatch-eql-alpha (fgf-eql :alpha) (:alpha))
+(test dispatch-eql-beta (fgf-eql :beta) (:beta))
+;;; Without memoization every eql-specialized call is a full dispatch miss.
+(test-true dispatch-eql-memoized
+           (progn (fgf-eql :alpha)
+                  (plusp (length (clos::generic-function-call-history #'fgf-eql)))))
--- a/src/lisp/regression-tests/mop.lisp
+++ b/src/lisp/regression-tests/mop.lisp
@@ -110,3 +110,20 @@
         (loop for n in nonwriters
               when (fboundp `(setf ,n))
                 collect n)))
+
+;;; AMOP allows a standard-class and a funcallable-standard-class in each
+;;; other's superclass chain; validate-superclass must accept both directions.
+(defclass vsc-plain () ((a :initform 1)))
+
+(test-true validate-superclass-mixed-metaclasses
+           (and (clos:validate-superclass (find-class 'standard-generic-function)
+                                          (find-class 'vsc-plain))
+                (clos:validate-superclass (find-class 'vsc-plain)
+                                          (find-class 'standard-generic-function))
+                t))
+
+(test-true validate-superclass-funcallable-from-standard
+           (progn (eval '(defclass vsc-funcallable (vsc-plain) ()
+                           (:metaclass clos:funcallable-standard-class)))
+                  (eq (class-of (find-class 'vsc-funcallable))
+                      (find-class 'clos:funcallable-standard-class))))


dg1sbg:
Note on the red `clasp/ubuntu-latest/bytecode/no/no` job on run [30693073178](https://github.com/clasp-developers/clasp/actions/runs/30693073178): it is **not** caused by this change. It failed on `WEAK-KEY-AND-VALUE-WEAKNESS`, a conservative-GC weak-hash-table test, which I have filed separately as #1814.

Evidence it is flaky rather than caused by this PR:

- The **identical commit** `661438586` re-run on a fork ([30697205345](https://github.com/dg1sbg/clasp/actions/runs/30697205345)) **passed** the regression-test step — 1968 successes, no failures. That job's red mark is an unrelated runner shutdown during the later ANSI step.
- In the same upstream run, `clasp/macos-latest/bytecode` and `clasp/ubuntu-latest/native` both passed with this commit. Only the Linux+bytecode intersection failed.
- The failing test has no causal connection to class-identity guards. I checked the one mechanism by which this PR could plausibly matter — the now-populated call histories retaining objects — and ruled it out: the eql-specializer intern table is unchanged by ordinary hash-table use (550 → 550), and a weak table still collapses to 1 after 500 EQL dispatches.

Local verification of this branch, macOS arm64:

- boehm **1977** successes, boehmprecise **1979**, zero unexpected failures on both
- a full from-scratch rebuild (`ninja -t clean`, 542/542 steps, 2.0 GB regenerated) then **1979** again
- `WEAK-KEY-AND-VALUE-WEAKNESS` specifically: 200/200 in isolation, 5/5 after the CLOS suites

I have re-triggered CI on this PR.

dg1sbg:
Second CI flake on this branch, a different one, same conclusion.

The re-triggered run cleared the earlier `WEAK-KEY-AND-VALUE-WEAKNESS` failure — `clasp/ubuntu-latest/bytecode` now passes. This time `clasp/macos-latest/bytecode` failed, at the ANSI step, on `UPGRADED-COMPLEX-PART-TYPE.9`: 1 unexpected failure out of 21936 tests, alongside 4 *unexpected successes* (`PRINT.LONG-FLOAT.RANDOM`, `PRINT.SYMBOL.RANDOM.3/4`, `FORMAT.E.26`).

Same commit `661438586`, same job, opposite results:

| run | `clasp/macos-latest/bytecode` |
|---|---|
| [30693073178](https://github.com/clasp-developers/clasp/actions/runs/30693073178) | success |
| [30701956588](https://github.com/clasp-developers/clasp/actions/runs/30701956588) | failure |

No commit in between — the second run was a close/reopen re-trigger of the identical SHA.

So across three runs of this branch, three different jobs failed for three unrelated reasons (weak-hash GC test, runner shutdown, one ANSI numeric test), while every other job passed each time. Nothing points at this change, which touches only CLOS class-identity guards. The four unexpected successes in the same ANSI run suggest that expected-failure list has drifted on macOS as well.

Local state for this branch remains: boehm **1977**, boehmprecise **1979**, zero unexpected failures, including a full from-scratch rebuild.

See #1814 for the weak-hash-table flake, which #1815 fixes.

dg1sbg:
Cross-platform validation, since this PR's CI has been red twice for reasons unrelated to it (see the two comments above — a weak-hash GC flake and an ANSI flake, each disproven by identical-commit contradiction).

Built and tested on a second platform to remove any doubt that the change is macOS/arm64-specific:

| | macOS arm64 / LLVM 22 / native | x86-64 Linux / LLVM 18 / bytecode |
|---|---|---|
| build | 542/542, clean from scratch | 536/536, clean from scratch |
| suite | **1979 successes, 0 unexpected failures** | **1979 successes, 0 unexpected failures** |
| expected failures | the same 4 | the same 4 |
| runs | full from-scratch rebuild + suite | **13 consecutive full suite runs, all clean** |

Same totals on both, across two architectures, two LLVM major versions, and two build modes. The Linux configuration is deliberately the same one whose CI job went red (`bytecode` / `no-default-native`).

Tested here as part of an integration branch merging six open PRs together, so this also confirms these three guards coexist with the shmem, slot-value, FFI and hardening work rather than only passing in isolation.

The six tests added by this PR pass on both platforms:

```
DISPATCH-EQL-ALPHA / -BETA / -MEMOIZED
VALIDATE-SUPERCLASS-MIXED-METACLASSES
VALIDATE-SUPERCLASS-FUNCALLABLE-FROM-STANDARD
```

Root cause of the underlying compiler defect is now fully diagnosed in #1811 — it is not `#.`-specific: `(eq X <constant instance>)` folds to `NIL` whenever X carries a CLOS class type declaration. This PR remains the correct workaround for the three kernel sites regardless of how that is fixed.

dg1sbg:
**Merging #1815 first is what makes this PR's CI able to hold still.**

The two red jobs here are not going to clear on their own, and re-triggering is close to a coin flip. Across three runs of this branch, three *different* jobs failed for three unrelated reasons, while every other job passed each time:

| run | failing job | cause |
|---|---|---|
| [30693073178](https://github.com/clasp-developers/clasp/actions/runs/30693073178) | `clasp/ubuntu-latest/bytecode` | `WEAK-KEY-AND-VALUE-WEAKNESS` |
| [30697205345](https://github.com/dg1sbg/clasp/actions/runs/30697205345) (fork, same SHA) | `clasp/ubuntu-latest/bytecode` | runner shutdown during ANSI step |
| [30701956588](https://github.com/clasp-developers/clasp/actions/runs/30701956588) | `cando/ubuntu-latest/native` | `WEAK-KEY-AND-VALUE-WEAKNESS` again |
| same run | `clasp/macos-latest/bytecode` | ANSI `UPGRADED-COMPLEX-PART-TYPE.9` |

Every one of those is disproven as a regression by identical-commit contradiction — the same SHA `661438586` passed each of those jobs on another run, with no commit in between.

`WEAK-KEY-AND-VALUE-WEAKNESS` accounts for two of the four, has now hit **two different jobs**, and is exactly what #1815 fixes (see #1814). It is not in this branch, so this PR cannot benefit from the fix while it sits unmerged.

Suggested order:

1. **#1815** — removes the dominant flake. Independent of this PR, one test file, no runtime change.
2. **#1813** — currently 6/6 green, independent, mergeable now.
3. **this PR** — re-trigger against a CI that can actually hold still.

Merging this one first would work too, but its CI would stay unreliable and every future re-trigger keeps rolling the same dice. There is nothing to fix in this PR's code: it passes 1979 successes with zero unexpected failures on macOS arm64 / LLVM 22 / native *and* x86-64 Linux / LLVM 18 / bytecode, the latter across 13 consecutive full suite runs in the same configuration whose CI job goes red.

dg1sbg:
I verified this PR on x86-64 Linux (LLVM 18, `boehmprecise`) against an `upstream/main` baseline, in **both** build modes. Both headline fixes are real and measurable — but the third change regresses the bytecode build, so I'd suggest splitting this PR.

## Both headline claims confirmed, with baseline contrast

`--build-mode=native`, where the read-time-guard miscompile actually bites. EQL-gf probe uses 30 EQL methods plus a default; `validate-superclass` probe uses the real case with **no** user-defined `validate-superclass` method (defining one masks the built-in guard entirely).

| | `upstream/main` (fc2beb475) | this PR (661438586) |
|---|---|---|
| EQL-gf call history | **0** — never memoized | **31** |
| bytes/call, first key | **16147.2** | **0.0** |
| bytes/call, last key | **16152.0** | **0.0** |
| ns/call | **113591** | **178.3** |
| `funcallable-standard-class` over `standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| `standard-class` over `funcallable-standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| genuinely illegal metaclass mix (negative control) | REJECTED | REJECTED |

`history=0` is categorical evidence the memoize guard was dead — the call history never populated, so every call took the full dispatch-miss path. That's a **637x** speedup and 16 KB -> 0 B per call. Note the cost is *flat* across first and last key (16147.2 vs 16152.0, 0.03% apart), which is the signature of a raw miss rather than a search.

The `validate-superclass` change is a genuine correctness fix: both AMOP-legal metaclass directions are rejected on `main`. The negative control still rejects an illegal mix on every arm, so the fix does not over-apply.

## The problem: `static-gfs/svuc.lisp` costs 122 tests in bytecode mode

`--build-mode=bytecode`, full regression suite:

| tree | successes | `CMPLTV::%INDEX` errors | aborted compilation units |
|---|---|---|---|
| `upstream/main` | 1963 | 0 | 2 (both expected) |
| **this PR as-is** | **1846** | **6** | **3** |
| **this PR minus `svuc.lisp`** | **1968** | **0** | **2** |

The failure is a `compile-file` abort:

```
; caught ERROR:
;   The slot CMPLTV::%INDEX in the object
;   #<CMPLTV::BASE-STRING-CREATOR ...> is unbound
; compilation unit aborted
```

around `SEQUENCES01.LISP`, so `ELT-*`, `EQUALP-*`, `ASSOC-*`, `BUTLAST-*` and
`CAN-MAP-TO-SPECIALIZED-VECTORS-*` never register. 1968 = 1963 baseline + the 5 tests this PR adds, so reverting that one file accounts for the difference exactly.

**This does not show up as a red build.** `ninja test` exits 0, because the harness fails on unexpected *failures*, not on tests that never registered — and nothing compares test counts. Four CI jobs on this PR are green while ~6% of the suite silently does not run. (Filing that separately.)

It is **bytecode-only** — same commit, `--build-mode=native`: 1968 successes, 0 CMPLTV errors. `CMPLTV` is the literal serializer for bytecode FASLs, and native uses a different literal path.

## Suggested split

Reverting **only** `src/lisp/kernel/clos/static-gfs/svuc.lisp` — keeping `miss.lisp` and `class.lisp` — was measured as a separate arm:

- both headline claims **identical** to the full PR (`history=31`, 0.0 B, both directions ACCEPTED)
- bytecode suite **1968 / 0 CMPLTV errors**
- native suite **1968 / 0 CMPLTV errors**
- the 5 new tests still pass

So `miss.lisp` + `class.lisp` deliver everything this PR claims, with a better suite result than baseline in both modes. `svuc.lisp` contributes to neither claim and is the sole cause of the regression.

One caveat on my part: I established **which** file causes it, not **why**. The plausible chain is that `uncustomizable-slot-p` now returning true routes through `gen-fast-value-read`, which for a `:class`-allocated slot emits a `load-time-value` wrapping `slotd-form` — and that form contains a string literal, which is what makes CMPLTV build a `BASE-STRING-CREATOR`. But a minimal reproducer built from that theory did **not** trigger the error, so the mechanism is unconfirmed and the underlying CMPLTV bug may well be pre-existing and merely made reachable here.

Happy to open the reduced PR if that split looks right to you.

dg1sbg:
**Retracting the split recommendation in my previous comment — the regression I reported does not reproduce, and the finding was an artifact of my own test methodology.**

## What I got wrong

I reported that this PR silently drops 122 regression tests in `--build-mode=bytecode` (1846 successes vs 1963 baseline), and attributed it to `static-gfs/svuc.lisp`. I called that a controlled single-variable experiment. It was not: it compared **one** run of one build against **one** run of a *differently produced* build, so build-to-build variation and change-under-test were confounded.

Re-testing properly — same commit `661438586`, `svuc.lisp` present and unmodified, three consecutive full-suite runs on a clean build:

| run | successes | aborted units | `CMPLTV::%INDEX` errors |
|---|---|---|---|
| 1 | 1968 | 2 (both expected) | 0 |
| 2 | 1968 | 2 | 0 |
| 3 | 1968 | 2 | 0 |

1968 = 1963 baseline + the 5 tests this PR adds. That is exactly the correct number. I also ran three further runs with an unrelated patch applied and got 1968 in all three, i.e. 6/6 healthy.

**So there is no 122-test regression, `svuc.lisp` is not implicated, and this PR does not need to be split.** I'm sorry for the noise.

## Probable cause of the bogus reading

The original 1846 came from a sweep that switched branches through a single reused build tree. That sweep predated my discovery of #1823 — koga's `generate-lisp-info` opens its eight outputs with `:if-exists :overwrite`, which does not truncate, so a regeneration producing *shorter* content leaves a stale tail on the previous file. The sweep order was #1809 -> this PR; #1809 adds files and symbols and this PR is based on an older commit with fewer, so the generated files shrink at exactly that transition — the precondition for the bug.

Every build since I found #1823 clears those files first, and the failure has not recurred. I'm running a direct test of that explanation now (same switch with and without clearing) and will follow up. If it confirms, it means #1823 can produce a *successful* build with a subtly wrong image rather than only a hard reader error, which makes that issue considerably more serious than I described it there — and I'll amend #1823 accordingly.

## What still stands

The verification of this PR's two headline fixes is unaffected — that was a separate three-arm native run with a proper baseline, not the bytecode sweep. On x86-64 Linux, `--build-mode=native`:

| | `upstream/main` (fc2beb475) | this PR |
|---|---|---|
| EQL-gf call history | **0** — never memoized | **31** |
| bytes/call, first key | **16147.2** | **0.0** |
| bytes/call, last key | **16152.0** | **0.0** |
| ns/call | **113591** | **178.3** |
| `funcallable-standard-class` over `standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| `standard-class` over `funcallable-standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| illegal metaclass mix (negative control) | REJECTED | REJECTED |

`history=0` on baseline is categorical evidence the memoize guard was dead — every call took the full dispatch-miss path, at 16 KB and 113 us. That is a **637x** speedup, and the cost is flat across first and last key (0.03% apart), which is the raw-miss signature rather than a search. The `validate-superclass` change fixes a genuine correctness bug in both metaclass directions, and the negative control confirms it does not over-apply.

Full suite on this PR: **1968 / `TEST_EXIT=0`** in `native` *and* `bytecode`.

Net: both claims verified on Linux with a proper baseline, no regression, **no split needed** — my earlier recommendation should be disregarded.

dg1sbg:
**Summary of my Linux verification of this PR. This supersedes my earlier comments in this thread (now collapsed) — one of them reported a regression that turned out not to reproduce, and I withdraw it.**

## Both fixes verified, with a baseline

x86-64 Linux, LLVM 18, `boehmprecise`, `--build-mode=native` — the mode where the read-time-guard miscompile actually bites. The EQL-gf probe uses 30 EQL methods plus a default; the `validate-superclass` probe uses the real case with **no** user-defined `validate-superclass` method, since defining one masks the built-in guard entirely.

| | `upstream/main` (fc2beb475) | this PR (661438586) |
|---|---|---|
| EQL-gf call history | **0** — never memoized | **31** |
| bytes/call, first key | **16147.2** | **0.0** |
| bytes/call, last key | **16152.0** | **0.0** |
| ns/call | **113591** | **178.3** |
| `funcallable-standard-class` over `standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| `standard-class` over `funcallable-standard-class` | **REJECTED** (`SIMPLE-ERROR`) | **ACCEPTED** |
| genuinely illegal metaclass mix (negative control) | REJECTED | REJECTED |

`history=0` on baseline is categorical evidence the memoize guard was dead — the call history never populated, so every call took the full dispatch-miss path at 16 KB and 113 us. That is a **637x** speedup and 16 KB -> 0 B per call. The cost is *flat* across first and last key (16147.2 vs 16152.0, 0.03% apart), which is the signature of a raw miss rather than a search.

The `validate-superclass` change fixes a genuine correctness bug — both AMOP-legal metaclass directions are rejected on `main`. The negative control still rejects an illegal mix, so the fix does not over-apply.

## Regression suite: clean

**1968 successes / `TEST_EXIT=0`** in `--build-mode=native` *and* `--build-mode=bytecode`. 1968 = 1963 baseline + the 5 tests this PR adds, which is exactly right. That figure held across nine runs on independently produced builds.

## Withdrawn: my earlier "drops 122 tests, split the PR" report

I earlier reported 1846 successes in bytecode mode and attributed it to `static-gfs/svuc.lisp`, recommending the PR be split. **That does not reproduce and the recommendation should be disregarded.** The reading came from a single run of a single build, compared against a single run of a differently produced build — build provenance and the change under test were confounded, so it was not the controlled comparison I described.

I chased three candidate explanations for the anomalous 1846 and none survived: `svuc.lisp` (the PR measures 1968 with it present and untouched), an unguarded slot read in `cmpltv`'s `print-object`, and stale generated files from #1823. For the last one I reproduced the exact conditions — build `feat/posix-shmem`, switch here without clearing the generated files, leaving 178 B and 214 B of stale tail on `runtime-packages.lisp` and `runtime-functions.lisp` — and the suite still came back 1968.

So the original 1846 is unreproduced and unexplained. I am not going to offer a fourth theory. Nothing here should block this PR.

## Two unrelated observations, `src/lisp/kernel/cmp/cmpltv.lisp`

Noted while investigating; neither is a fix for any bug I can currently demonstrate:

- `print-object` for `vcreator` guards `%prototype` with `slot-boundp` but reads `(index object)` unguarded. Any half-constructed creator therefore raises a secondary unbound-slot error *from inside the printer*, masking whatever the real condition was — which is exactly how the original failure presented and a large part of why it was hard to read.
- The `%index` slot is declared `:type (integer 0)` while its initform is `nil`; upstream Maclina has `:type (or null (integer 0))`. Not a live bug (`make-instance` stores `nil` there without complaint), but the declaration contradicts the initform.

Happy to send a small PR for those if useful.

dg1sbg:
Small correction to the byte counts in my previous comment (the conclusions are unchanged).

I listed this PR's clean generated-file sizes as 88034 / 70658. Those figures are from a different branch measured in an earlier session — I carried them over without re-deriving them. Clearing the files and rebuilding at this commit gives the true sizes:

| file | #1809 clean | after dirty rebuild here | true size here | stale tail |
|---|---|---|---|---|
| `runtime-packages.lisp` | 88561 | 88561 | 88383 | 178 B |
| `runtime-functions.lisp` | 71216 | 71216 | 71002 | 214 B |

so the tail was 178 / 214 bytes, not the 527 / 558 I quoted.

Everything else stands: the dirty build still succeeded and its suite still came back **1968 / TEST_EXIT=0 / 0 CMPLTV errors**, so #1823 does not explain the original anomalous reading either. This PR measures healthy at 1968 in every run, and the 1846 remains unreproduced and unexplained.

PR #1815, opened 2026-08-01
tests: make weak-table weakness tests robust under conservative GC

### LINKED ISSUE #1814: Flaky test: WEAK-KEY-AND-VALUE-WEAKNESS fails intermittently on ubuntu-latest/bytecode
`WEAK-KEY-AND-VALUE-WEAKNESS` (`src/lisp/regression-tests/hash-tables0.lisp:147`) fails intermittently in CI. Same commit, same job, opposite results:

| run | commit | `clasp/ubuntu-latest/bytecode/no/no` |
|---|---|---|
| clasp-developers/clasp [30693073178](https://github.com/clasp-developers/clasp/actions/runs/30693073178) | `661438586` | "Run regression tests" **failed** — `Failures: WEAK-KEY-AND-VALUE-WEAKNESS`, Successes: 1967 |
| dg1sbg/clasp [30697205345](https://github.com/dg1sbg/clasp/actions/runs/30697205345) | `661438586` (identical) | "Run regression tests" **passed** — Successes: 1968 |

(The fork run's job still shows red, but for an unrelated reason: the runner received a shutdown signal during the later ANSI step. Its regression-test step succeeded.)

Only this OS/mode combination is affected. In the upstream run above, `clasp/macos-latest/bytecode` and `clasp/ubuntu-latest/native` both passed with the same commit.

## Why it is fragile

```lisp
(test weak-key-and-value-weakness
      (let ((table (make-hash-table :weakness :key-and-value)))
        (setf (gethash (list nil) table) :value
              (gethash :key table) (list nil)
              (gethash (list nil) table) (list nil)
              (gethash :key table) :value)
        (gctools:garbage-collect)
        (hash-table-count table))
      (1))
```

The test allocates three throwaway `(list nil)` values, triggers one GC, and asserts the count is *exactly* 1. Under Boehm's conservative stack scanning, a single stale word in a register or stack slot still referencing one of those temporaries keeps its entry alive and the count comes out 2. Whether that happens depends on register allocation and stack residue at the call site, so it is sensitive to unrelated codegen changes and to the state left behind by the ~20 suites that run before `hash-tables0`.

`weak-key-weakness` immediately above it has the same shape and presumably the same exposure.

## Not reproducible locally

macOS arm64, both variants, all clean:

- 200/200 iterations of the test body in a fresh image
- 5/5 runs of `TEST_SUITES=fastgf,clos,mop,hash-tables0`
- full suites: boehm 1977, boehmprecise 1979, zero unexpected failures, including a from-scratch rebuild

## Suggestion

Rather than asserting an exact count after one GC, either assert the surviving entry is present (`(gethash :key table)` → `:value`) and treat extra entries as tolerable, or retry the collect a bounded number of times before asserting. As written the test can redden any PR at random, which makes red CI easy to dismiss — the failure mode you least want on a project where CI is the stability gate.

Found while working on #1811 / #1812; unrelated to that change, which touches only CLOS class-identity guards and creates no new retention path (verified: the eql-specializer intern table is unchanged by hash-table use, and weak tables still collapse to 1 after 500 EQL dispatches with fully populated call histories).

dg1sbg:
Fixes #1814.

`WEAK-KEY-AND-VALUE-WEAKNESS` fails intermittently on `clasp/ubuntu-latest/bytecode`. The **identical commit** failed the regression step on one run and passed it on another, while `macos/bytecode` and `ubuntu/native` passed both times — so it is non-deterministic, not a real regression. Details and run links in #1814.

## Why the tests are fragile

All four weakness tests share this shape:

```lisp
(let ((table (make-hash-table :weakness :key-and-value)))
  (setf (gethash (list nil) table) :value
        ...)
  (gctools:garbage-collect)
  (hash-table-count table))
```

Two things make that unreliable under Boehm's conservative stack scanning:

1. The throwaway `(list nil)` values are created in **the same frame that then calls `garbage-collect`**, so their addresses are still sitting in live stack slots and registers when the collector scans the stack. The collector cannot tell those words from real references and keeps the entries alive.
2. A single collection is assumed to be enough.

Either one leaves a dead entry alive and the count comes out one too high. Because it depends on register allocation and stack residue, it can be perturbed by entirely unrelated codegen changes elsewhere in the tree — which is how it surfaced.

The existing comment already acknowledged these are "pretty strict"; this makes them strict about the property being tested rather than about incidental stack layout.

## Change

- Populate the table in a builder function that **returns before** the collection, so the frame holding the temporaries is popped and below the stack pointer when the scan happens.
- Collect up to `TRIES` times rather than exactly once.
- One `WEAK-COUNT` helper shared by all four tests instead of repeating the retry logic four times.

## Not vacuous

A retry loop risks turning the test into a tautology, so I checked that it still fails on genuine retention. Holding one key in a global:

```
dead temporaries      => 1   (expected 1, passes)
one key held globally => 2   (expected 1, still fails after all attempts)
```

An entry that is really retained never reaches the expected count, so the tests keep their teeth.

## Verification

macOS arm64, boehmprecise: full suite **1979 successes, zero unexpected failures**; the weakness tests pass 3/3 across repeated `TEST_SUITES=hash-tables0` runs. (`weak-key-or-value-weakness` is `#-use-boehm` and does not run on these variants, but is updated for consistency.)

### DIFF
--- a/src/lisp/regression-tests/hash-tables0.lisp
+++ b/src/lisp/regression-tests/hash-tables0.lisp
@@ -127,41 +127,60 @@
 
 ;;; These tests are pretty strict - they want the garbage to actually be
 ;;; collected by that garbage-collect call, which may not be the case if we
-;;; ever get a more relaxed collector (generational or something)
+;;; ever get a more relaxed collector (generational or something).
+;;;
+;;; BUILDER is called to populate the table and then returns, so the frame
+;;; holding the dead temporaries is popped before we collect - otherwise
+;;; conservative stack scanning can see a stale reference and keep an entry
+;;; alive. For the same reason we collect up to TRIES times instead of
+;;; demanding the exact count after a single collection: a genuinely retained
+;;; entry still fails, but a transient stack artifact does not.
+(defun weak-count (builder expected &optional (tries 10))
+  (let ((table (funcall builder)))
+    (loop repeat tries
+          do (gctools:garbage-collect)
+          when (eql (hash-table-count table) expected)
+            return expected
+          finally (return (hash-table-count table)))))
+
 (test weak-key-weakness
-      (let ((table (make-hash-table :weakness :key)))
-        (setf (gethash (list 37) table) :value
-              (gethash :key table) (list nil)
-              (gethash :key2 table) (list nil))
-        (gctools:garbage-collect)
-        (hash-table-count table))
+      (weak-count (lambda ()
+                    (let ((table (make-hash-table :weakness :key)))
+                      (setf (gethash (list 37) table) :value
+                            (gethash :key table) (list nil)
+                            (gethash :key2 table) (list nil))
+                      table))
+                  2)
       (2))
 (test weak-value-weakness
-      (let ((table (make-hash-table :weakness :value)))
-        (setf (gethash :key table) (list 37)
-              (gethash (list nil) table) :value
-              (gethash (list 18) table) :value)
-        (gctools:garbage-collect)
-        (hash-table-count table))
+      (weak-count (lambda ()
+                    (let ((table (make-hash-table :weakness :value)))
+                      (setf (gethash :key table) (list 37)
+                            (gethash (list nil) table) :value
+                            (gethash (list 18) table) :value)
+                      table))
+                  2)
       (2))
 (test weak-key-and-value-weakness
-      (let ((table (make-hash-table :weakness :key-and-value)))
-        (setf (gethash (list nil) table) :value
-              (gethash :key table) (list nil)
-              (gethash (list nil) table) (list nil)
-              (gethash :key table) :value)
-        (gctools:garbage-collect)
-        (hash-table-count table))
+      (weak-count (lambda ()
+                    (let ((table (make-hash-table :weakness :key-and-value)))
+                      (setf (gethash (list nil) table) :value
+                            (gethash :key table) (list nil)
+                            (gethash (list nil) table) (list nil)
+                            (gethash :key table) :value)
+                      table))
+                  1)
       (1))
 #-use-boehm ; on boehm key-or-value tables are effectively strong.
 (test weak-key-or-value-weakness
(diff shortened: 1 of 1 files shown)


dg1sbg:
Tried to validate this on Linux, where the flake actually occurs. **The result is inconclusive, and I would rather say so than imply the fix is proven.**

Setup: x86-64 Ubuntu 24.04, LLVM 18, `--build-mode=bytecode --no-default-native` — i.e. the same configuration as the `clasp/ubuntu-latest/bytecode/no/no` job that failed twice on `WEAK-KEY-AND-VALUE-WEAKNESS`. Two arms, same machine, same binary, differing only in which version of `hash-tables0.lisp` the harness compiles at run time:

| arm | runs | `WEAK-*` failures |
|---|---|---|
| with this fix | 6 | 0 |
| **with this fix reverted (control)** | **6** | **0** |

The control passed too. So this box never reproduced the flake, and the six green runs on the fixed side establish nothing about the fix. Reproduction likely needs CI's environment — 2-4 core runners under memory pressure have a very different stack and heap profile than the 16-core / 157 GB box I used, and conservative-GC retention is exactly the sort of thing that depends on that.

What *is* established, independent of reproduction:

- The hazard is real. The dead `(list nil)` temporaries are created in the same frame that then calls `garbage-collect`, so their addresses are still in live stack slots when the collector scans. Building them in a function that returns first puts that frame below the stack pointer.
- The change does not make the tests vacuous. Holding one key in a global keeps the count at 2 through every retry, so genuine retention still fails.
- No regression: full suites pass on macOS arm64 / LLVM 22 / native (1979 successes, 0 unexpected failures) and on x86-64 Linux / LLVM 18 / bytecode (1979, 0 unexpected failures, 13 consecutive runs).

So this rests on the reasoning plus the non-vacuity check, not on an observed before/after. If the flake reappears after this merges, that is the reason.

Bike:
For the moment I'm just going to mark the tests as known failures under Boehm. Boehm is not reliable enough for us to be able to mandate that objects are collectable.

PR #1842, opened 2026-08-20
Expect the randomized ~E tests FORMAT.E.1 and FORMAT.E.2 to fail sometimes

### LINKED ISSUE #1816: prin1 prints some double-floats with too few digits to read back the same float

dg1sbg:
`FORMAT.E.1` and `FORMAT.E.2` fail intermittently on unmodified `main` and on unrelated pull requests, and each failure costs someone an adjudication before the red can be dismissed. They are not deterministic `~E` tests. Each draws a float type with `(RANDOM 4)`, draws a random exponent, repeats 1000 times, and compares `(FORMAT NIL "~E" X)` against `PRIN1-TO-STRING` and against a compiled `FORMATTER` call (`dependencies/ansi-test/printer/format/format-e.lsp:12,45`). Whether they trip is a property of the draw.

What they catch when they do trip is the round-trip float-printing defect already tracked as #1816, reached through `~E` rather than through `PRIN1`: the two printers disagree about where the decade boundary falls when the mantissa rounds up to it. On a build of `0aa5b71ac`, `~E` prints the single-float `9.99999e-18` as `"0.999999e-17"` while `PRIN1` prints `"9.99999e-18"`, and the double-float `9.9999999e-18` disagrees the same way. It is not specific to `LONG-FLOAT` and not specific to extreme exponents; any of the four float types will do, at ordinary magnitudes.

Recent instances: `main` at `0aa5b71ac` red on `FORMAT.E.1` with that exact single-float value, and #1820 red on `FORMAT.E.2` with `-9.999338424609615l-2430`. In both cases it was the only unexpected failure out of 21936, with the regression suite green in the same job.

### Why listing them is free

`DO-ENTRIES` returns `(null *unexpected-failures*)` (`rt.lsp:521`), so a listed test that passes is recorded as an unexpected success (`rt.lsp:496`) and reported, but does not change the exit status. Both names match loaded test entries, so neither adds to the `Found unknown test or note names` report — verified by loading `init.lsp` and calling `rt::load-expected-failures` against the edited file:

```
UNKNOWN-EXPECTED-FAILURES: NIL
FORMAT.E.1: symbol=T known-test=T listed-as-expected-failure=T
FORMAT.E.2: symbol=T known-test=T listed-as-expected-failure=T
TOTAL-EXPECTED-FAILURES: 32
```

This is already how `PRINT.LONG-FLOAT.RANDOM`, `PRINT.SYMBOL.RANDOM.3`, `PRINT.SYMBOL.RANDOM.4` and `FORMAT.E.26` are handled; `FORMAT.E.26` is defined in the same file and already carries the `fails sometimes` note.

### Scope

This does not fix #1816, and it deliberately leaves `PRINT.SHORT-FLOAT.RANDOM`, `PRINT.SINGLE-FLOAT.RANDOM` and `PRINT.DOUBLE-FLOAT.RANDOM` unlisted even though `PRINT.DOUBLE-FLOAT.RANDOM` has also been seen red on `main`. Those are a judgement for whoever owns the printer work; the smallest change that stops the recurring noise is the one worth making now. Happy to widen it if you would rather have the whole randomized-printer family listed at once.

Refs #1816


dg1sbg:
Widened, and the CI run on the first version is the reason.

Run [32345271440](https://github.com/clasp-developers/clasp/actions/runs/32345271440), job [96352593574](https://github.com/clasp-developers/clasp/actions/runs/32345271440/job/96352593574) — `clasp/ubuntu-latest/native/yes/no`, step `Run ANSI tests`:

```
29 failures with 1 unexpected failure and 4 unexpected successes out of 21936 tests.
Unexpected failures:
  PRINT.DOUBLE-FLOAT.RANDOM
Unexpected successes:
  PRINT.SYMBOL.RANDOM.3
  PRINT.SYMBOL.RANDOM.4
  FORMAT.E.1
  FORMAT.E.2
FAILED: [code=1] ansi-test-boehmprecise
```

That single job contains both halves of the argument. `FORMAT.E.1` and `FORMAT.E.2` drew clean, landed in *unexpected successes*, and contributed nothing to the exit status — the mechanism works exactly as described. And the job still went red, on `PRINT.DOUBLE-FLOAT.RANDOM`: the test I had deliberately left out and flagged in the description as also known to red on `main`. The next draw from the same family reddened the next cell.

So this now also lists `PRINT.SHORT-FLOAT.RANDOM`, `PRINT.SINGLE-FLOAT.RANDOM` and `PRINT.DOUBLE-FLOAT.RANDOM`, joining `PRINT.LONG-FLOAT.RANDOM`, which was already there. Those four are the same test written once per float type — each walks the exponent range from least-positive to most-positive, draws ten values per decade with `(RANDOM f)` and feeds them to `RANDOMLY-CHECK-READABILITY` (`print-floats.lsp:86,181,276,371`, checking at `:100,195,290,385`). Listing one of the four and not the other three was arbitrary; the underlying round-trip defect (#1816) is not specific to a float type.

Re-verified against the loaded test entries before pushing, same way as before — no unknown names, 32 → 35 expected failures:

```
UNKNOWN-EXPECTED-FAILURES: NIL
PRINT.SHORT-FLOAT.RANDOM:  known-test=T listed=T
PRINT.SINGLE-FLOAT.RANDOM: known-test=T listed=T
PRINT.DOUBLE-FLOAT.RANDOM: known-test=T listed=T
PRINT.LONG-FLOAT.RANDOM:   known-test=T listed=T
FORMAT.E.1:                known-test=T listed=T
FORMAT.E.2:                known-test=T listed=T
TOTAL-EXPECTED-FAILURES: 35
```

Still no attempt at #1816 itself — this only stops the randomized printer tests from reddening unrelated PRs. Happy to trim it back if you would rather keep some of these visible.

PR #1854, opened 2026-09-09
regression-tests: byte-exact allocation suites for both tiers

dg1sbg:
Two byte-exact allocation suites, one per tier, plus the LOOP rows. No behaviour change.

`allocation.lisp` — bytecode tier, every shape via `CMP:BYTECOMPILE`.
`allocation-native.lisp` — native tier, every shape via `CL:COMPILE` under `CMP:*COMPILE-NATIVE*`, and it **refuses a result whose entry is not a `SIMPLE-CORE-FUN`**, so a row cannot silently measure the other tier.
`allocation-helpers.lisp` — the sinks and CLOS fixture both share.

Each row divides total bytes by call count, so a partial result shows as a ratio, never a rounded 0. Controls first (`cons` reads 24, an empty call 0); constructs that still allocate are listed in `set-unexpected-failures.lisp` with today's figure in the `:description`.

Readings this pins down (native, per instance):

- escaping closure `32 + 8×captures`; a capture-free lambda is a literal, 0
- a captured variable with ≥2 writers: `+24` cell, at its **binding**
- `unwind-protect` cleanup 40 / 56 / 64; `mp:with-lock` 40
- `handler-bind` 72 (+48 per binding); `handler-case` 144 / 176; `ignore-errors` 176
- a method mentioning `call-next-method`: 24 **per required argument**, on every call
- `catch`/`throw`, `block`/`return-from`, special binding, capturing `multiple-value-bind`: 0
- a surplus mvb value: 24 under `(compile nil)`, 0 under `compile-file` — asserted as the pipeline fact it is

`loop.lisp` gains four rows for `for v of-type T = form` compiled natively (`cons`, with `then`, an array type, and a `fixnum` control) — red before s-expressionists/Khazern#18, green after.

`TEST_SUITES=allocation,allocation-native ninja test`: exit 0.

---

**Stack.** Base of the series: #1855, #1856 and #1857 build on this one and each carries it in its diff. #1853 is independent.

### DIFF
--- a/src/lisp/regression-tests/allocation-helpers.lisp
+++ b/src/lisp/regression-tests/allocation-helpers.lisp
@@ -1,0 +1,29 @@
+(in-package #:clasp-tests)
+
+;;;; The helpers the two allocation suites share: allocation.lisp (the bytecode tier) and
+;;;; allocation-native.lisp (the native tier) measure the same shapes through the same sinks, so the
+;;;; sinks live once, here, and each suite LOADs this file at its top.  Loading it twice is harmless.
+
+(defvar *alloc-sink* nil)
+(defvar *alloc-lock* (mp:make-lock :name "allocation-test"))
+
+(define-condition %alloc-full (error) ())
+
+(declaim (notinline %alloc-touch %alloc-two %alloc-three %alloc-global-handler %alloc-call))
+(defun %alloc-touch (x) (setq *alloc-sink* x) nil)
+(defun %alloc-two (i) (values i (1+ i)))
+(defun %alloc-three (i) (values i (1+ i) (+ i 2)))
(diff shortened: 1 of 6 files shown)


Bike:
I appreciate that you're putting in all these PRs, but are you not looking them over? I'm not going to put in a test that conses are exactly 24 bytes and so on. There could maybe be a place for no-consing tests but I'm obviously not going to mandate that conses are exactly 24 bytes, and closures are so many bytes, and etc. At best I'd put in a max size. And this PR doesn't actually explain what it's doing. "byte-exact allocation suite" is not very descriptive - this reads like you just copy pasted the second half of what the LLM told you.

